PR #2, opened 2020-09-08
P4-2136 reworking JPA entities to ensure validation is enforced.

matt-crowson-moj:


jcavell (approved):

jcavell:
After discussion about the @get annotation, looks good

PR #8, opened 2020-09-18
P4 2185 import reporting data

jcavell:
@matt-crowson-moj PR for importing the reporting files, not using JPA


matt-crowson-moj (approved):
LGTM

PR #19, opened 2020-09-29
P4-2260 changes to tighten and improve the validation around the prices import process.

matt-crowson-moj:
Changes to tighten up the validation when importing prices for the given suppliers.  No longer relying entirely on JPA constraints.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsSpreadsheet.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsSpreadsheet.kt
@@ -20,9 +20,8 @@ private const val ROW_OFFSET = 1
 class LocationsSpreadsheet(private val spreadsheet: Workbook, private val locationRepository: LocationRepository) : Closeable {
 
     init {
-        Tab.values().toList().filter { spreadsheet.getSheet(it.label) == null }.joinToString { it.label }.let {
-            if (it.isNotBlank()) throw IllegalStateException("The following tabs are missing from the locations spreadsheet: $it")
-        }
+        Tab.values().toList().filter { spreadsheet.getSheet(it.label) == null }.joinToString { it.label }
+                .takeIf { it.isNotBlank() }?.let { throw NullPointerException("The following tabs are missing from the locations spreadsheet: $it") }
     }
 
     enum class Tab(val label: String) {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsSpreadsheet.kt
```diff
@@ -20,9 +20,8 @@ private const val ROW_OFFSET = 1
 class LocationsSpreadsheet(private val spreadsheet: Workbook, private val locationRepository: LocationRepository) : Closeable {
 
     init {
-        Tab.values().toList().filter { spreadsheet.getSheet(it.label) == null }.joinToString { it.label }.let {
-            if (it.isNotBlank()) throw IllegalStateException("The following tabs are missing from the locations spreadsheet: $it")
-        }
+        Tab.values().toList().filter { spreadsheet.getSheet(it.label) == null }.joinToString { it.label }
```
matt-crowson-moj, line 23: @andymarke FYI

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsSpreadsheet.kt
```diff
@@ -20,9 +20,8 @@ private const val ROW_OFFSET = 1
...
+                .takeIf { it.isNotBlank() }?.let { throw NullPointerException("The following tabs are missing from the locations spreadsheet: $it") }
```
andymarke, line 24: sanity check; should not be it.isBlank? i.e throw exception only when blank?

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsSpreadsheet.kt
```diff
@@ -46,11 +45,11 @@ class LocationsSpreadsheet(private val spreadsheet: Workbook, private val locati
         val locationType = LocationType.map(cells[TYPE].stringCellValue.toUpperCase().trim())
                 ?: throw IllegalArgumentException("Unsupported location type: " + cells[TYPE].stringCellValue)
 
-        val agency = cells[AGENCY].stringCellValue.toUpperCase().trim().takeUnless { it == "" }
-                ?: throw IllegalArgumentException("Agency id cannot be blank")
+        val agency = cells[AGENCY].stringCellValue.toUpperCase().trim().takeUnless { it.isBlank() }
```
andymarke, line 48: Minor (personal view), I think I would either have used `!!` to force an exception on null or used an IllegalArgument/State type exception rather than throwing a NPE manually.

Though I'm sure you end up with the correct behaviour either way, so no big deal

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsSpreadsheet.kt
matt-crowson-moj: No. if the joined string is not blank then we have some tabs that are not present.

PR #25, opened 2020-10-01
p4-2186

jcavell:
ReportingImporter gets files between to and from dates
MoveFilterer returns moves with MOVE_START occurred_at between to and from dates

ImportController.kt now takes a query string for movesFrom, movesTo and ReportsTo ISO dates
ImportService.kt takes these new params and passes movesFrom/reportsTo to the reporting Importer.kt
ImportService.kt passes movesFrom/movesTo to the PriceCalculator's standardPrices method via nw MoveFiltererParams

Change ReportingImporter.kt to take in parameters for from and end dates for importing
Likewise, change fileNamesForDate to take the from and to dates

Refactor MoveFilterer.kt standardMoves and simpleRedirectMoves for better readability
Introduce MoveFiltererTest.kt

Remove pointless Importer interface


matt-crowson-moj (approved):
LGTM

PR #26, opened 2020-10-01
P4-2228 including from and to in standard moves spreadsheet and refactoring for reuse.

matt-crowson-moj:
Changes are predominantly refactoring so we can easily process other sheets on the template spreadsheet.

Also now includes the from and to in the outputted spreadsheet.

Got rid of some unused imports.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/output/PriceSheet.kt
```diff
@@ -0,0 +1,45 @@
...
+    private fun applyHeader() {
+        sheet.getRow(0).getCell(1).setCellValue(header.dateRun)
+        sheet.getRow(4).createCell(1).setCellValue(header.supplier.reportingName().capitalize())
+        sheet.getRow(5).getCell(1).setCellValue(header.dateRange.from)
+        sheet.getRow(5).getCell(3).setCellValue(header.dateRange.to)
+
+        // TODO need to add version as well.
+    }
+
+    fun add(prices: Sequence<MovePrice>) {
```
matt-crowson-moj, line 28: @jcavell not sure but we might need/want to make this generic???

jcavell (approved):

PR #29, opened 2020-10-02
p4-2186 add columns to standard moves sheet

jcavell:
Add date and time columns
Convenience method to get latest events of a type, on Event.kt


matt-crowson-moj (approved):
LGTM

PR #30, opened 2020-10-02
P4-2228 refactored the spreadsheet generator.  This now takes in the calculator so generate and the sheet level.

matt-crowson-moj:
This change pushes the calculation into the spreadsheet generator.  This allows us to calculate on a sheet by sheet basis instead of doing all up front and passing in, and helps to minimise work clashes as changes in the generator should now be small (e.g. adding a new sheet to be calculated).

Also extending the test coverage though as always there is room for improvement/more.

Added in mockito config into test resources.  By default Kotlin classes are final (unless open), this change means we can mock final classes.


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/TestConfig.kt
```diff
@@ -19,7 +20,7 @@ class TestConfig {
     private lateinit var resourceLoader: ResourceLoader
 
     @Bean
-    fun clock() = Clock.systemDefaultZone()
+    fun clock(): Clock = Clock.fixed(Instant.now(), ZoneId.systemDefault())
```
matt-crowson-moj, line 23: It is a test clock so ideally we want it fixed so we can be pretend time-lords.

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
matt-crowson-moj, line 1: This allows us to mock (final) Kotlin classes.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/output/PricesSpreadsheetGeneratorTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.time.Clock
+import java.time.LocalDate
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("test")
+@ContextConfiguration(classes = [TestConfig::class])
+internal class PricesSpreadsheetGeneratorTest(@Autowired @Qualifier(value = "spreadsheet-template") template: File, @Autowired private val clock: Clock) {
+
+    private val calculator: PriceCalculator = mock { onGeneric { standardPrices(any()) } doReturn sequenceOf() }
+    private val generator: PricesSpreadsheetGenerator = PricesSpreadsheetGenerator(template, clock)
```
matt-crowson-moj, line 24: Initial intent here was to wrap this with a Spy so I can interrogate the inner behaviour more but for some reason I couldn't get it to work.  I am probably missing something.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/output/PricesSpreadsheetGenerator.kt
```diff
@@ -1,25 +1,33 @@
...
 @Component
 class PricesSpreadsheetGenerator(@Autowired @Qualifier(value = "spreadsheet-template") private val template: File,
-                                 @Autowired private val clock: Clock) {
+                                          @Autowired private val clock: Clock) {
 
-    fun generateSpreadsheet(from: LocalDate, to: LocalDate, supplier: Supplier, prices: Sequence<MovePrice>): File {
-        val dateRun = LocalDate.now(clock)
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    internal fun generate(from: LocalDate, to: LocalDate, supplier: Supplier, calculator: PriceCalculator): File {
```
jcavell, line 23: from, to, supplier look like the same as in MoveFiltererParams, so we should probably use this instead. Likewise, we probably don't need a dedicated PriceSheet.DateRange, that could also be within the MoveFiltererParams (which we could rename to make more generic). We could use a type of ClosedRange<LocalDate> to represent the date range.

matt-crowson-moj: Yeah agreed.  The date range this is a bit of a hangover and doesn't add much.

matt-crowson-moj: Have added a ClosedRangeLocalDate see what you think.  Not doe a change to the MoveFiltererParams.  Perhaps leave/do separately as may get a clash.

jcavell (approved):

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
andymarke: @matt-crowson-moj I switched to including `testImplementation("org.mockito:mockito-inline:3.4.6")`

andymarke (commented):
LGTM

PR #31, opened 2020-10-02
Deploy config

mafonso:
First pass and helm config definitions.

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -8,6 +8,9 @@ image:
   tag: latest
   port: 8080
 
+db:
+  secret_name:
+
 ingress:
   enabled: true
   enable_whitelist: true
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -12,7 +12,7 @@ ingress:
   enabled: true
   enable_whitelist: true
   hosts:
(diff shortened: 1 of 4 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -1,18 +1,21 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-replicaCount: 2
+replicaCount: 1
```
matt-crowson-moj, line 4: Perhaps put a comment against this why setting to 1??  not a big deal if none but could potentially be an issue if changed.

helm_deploy/values-prod.yaml
```diff
@@ -1,7 +1,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-replicaCount: 4
+replicaCount: 1
```
matt-crowson-moj, line 4: Same as the dev comment for pre-prod and prod.  Very though.

matt-crowson-moj (approved):
LGTM

helm_deploy/values-dev.yaml
mafonso: Yes, can add that. 
This are per env values, that usually will be reviewed at different stages of the project.

PR #33, opened 2020-10-02
P4-2228 change on the back of comments to tidy up the call to the spreadsheet generate method.

matt-crowson-moj:
Tidy up the params on the Spreadsheet generator.

Also removed mockito test resource and replace with test dependency which does the same job.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,7 +17,9 @@ dependencies {
 	implementation("com.github.kittinunf.result:result:3.1.0")
 	implementation("com.github.kittinunf.result:result-coroutines:3.1.0")
 	implementation("com.beust:klaxon:5.4")
-	implementation("com.amazonaws:aws-java-sdk-s3:1.11.873")
+	implementation("com.amazonaws:aws-java-sdk-s3:1.11.874")
+
+	testImplementation("org.mockito:mockito-inline:3.5.13")
 
 	runtimeOnly("com.h2database:h2")
 	runtimeOnly("org.postgresql:postgresql:42.2.16")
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -17,7 +17,9 @@ dependencies {
 	implementation("com.github.kittinunf.result:result:3.1.0")
 	implementation("com.github.kittinunf.result:result-coroutines:3.1.0")
 	implementation("com.beust:klaxon:5.4")
-	implementation("com.amazonaws:aws-java-sdk-s3:1.11.873")
+	implementation("com.amazonaws:aws-java-sdk-s3:1.11.874")
+
+	testImplementation("org.mockito:mockito-inline:3.5.13")
```
matt-crowson-moj, line 22: This removes the need to have something specific in the test resource folder.  It enables mock of final classes (which most are in Kotlin).

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/reporting/MoveFilterer.kt
```diff
@@ -55,5 +56,7 @@ object MoveFilterer {
 data class MoveFiltererParams(
         val supplier: Supplier,
         val movesFrom: LocalDate,
-        val movesTo: LocalDate,
-)
\ No newline at end of file
+        val movesTo: LocalDate) {
+
+    fun dateRange() = ClosedRangeLocalDate(movesFrom, movesTo)
```
matt-crowson-moj, line 61: Added here to avoid changing the interface (constructor) and having rippled effect of changes elsewhere.

jcavell (approved):

PR #35, opened 2020-10-02
p4-2186 add columns to standard moves sheet

jcavell:
Not pretty, but functional

Create RowValue.kt to hold the display value for each cell
Add fromLocationType and toLocationType to MovePrice.kt
Create an addStandardRow fun on PriceSheet.kt which StandardMovesSheet uses

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/calculator/MovePrice.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/calculator/MovePrice.kt
@@ -1,9 +1,12 @@
 package uk.gov.justice.digital.hmpps.pecs.jpc.calculator
 
+import uk.gov.justice.digital.hmpps.pecs.jpc.location.LocationType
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/calculator/MovePrice.kt
```diff
@@ -1,9 +1,12 @@
 package uk.gov.justice.digital.hmpps.pecs.jpc.calculator
 
+import uk.gov.justice.digital.hmpps.pecs.jpc.location.LocationType
 import uk.gov.justice.digital.hmpps.pecs.jpc.reporting.JourneyWithEvents
 import uk.gov.justice.digital.hmpps.pecs.jpc.reporting.MovePersonJourneysEvents
 
 data class MovePrice(
+        val fromLocationType: LocationType?,
```
matt-crowson-moj, line 8: Is that same problem around some moves not having locations in dev?

matt-crowson-moj (approved):
LGTM

PR #38, opened 2020-10-05
Shared ingress

mafonso:
shared ingress

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -13,19 +13,11 @@ db:
 
 ingress:
   enabled: true
-  enable_whitelist: true
+  enable_whitelist: false
   hosts:
-    - host: calculate-journey-variable-payments-dev.hmpps.service.justice.gov.uk
-      cert_secret: calculate-journey-variable-payments-cert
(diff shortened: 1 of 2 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -13,19 +13,11 @@ db:
 
 ingress:
   enabled: true
-  enable_whitelist: true
+  enable_whitelist: false
   hosts:
-    - host: calculate-journey-variable-payments-dev.hmpps.service.justice.gov.uk
-      cert_secret: calculate-journey-variable-payments-cert
+    - host: calculate-journey-variable-payments-dev.apps.live-1.cloud-platform.service.justice.gov.uk
```
matt-crowson-moj, line 18: Wow that is a bit of a mouthful.  Is that right?

matt-crowson-moj (approved):
LGTM

helm_deploy/values-dev.yaml
mafonso: Yes, this one of the ways of doing it for non prod environments.  We can have a custom one, but is that really required for dev?

PR #44, opened 2020-10-06
P4-2228 change move and journey to have Location type rather than string

jcavell:
This is so that we can easily get the sitename and location type.

Achieved using LocationParser.kt which looks at a list of locations.
If the location name is null it returns null
If the location for the nomis agency id exists, it uses this
If the location doesn't exist, it creates a new one with the agency id as sitename and locationtype unknown.

Remove location mapping from PriceCalculator.kt as it's no longer needed.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/output/OutputSpreadsheet.kt
```diff
@@ -23,8 +23,8 @@ class OutputSpreadsheet {
 
             with(p.moveReport.move){
                 row.createCell(0).setCellValue(reference)
-                row.createCell(1).setCellValue(fromLocation)
-                row.createCell(2).setCellValue(toLocation)
+                row.createCell(1).setCellValue(fromLocation?.siteName)
+                row.createCell(2).setCellValue(toLocation?.siteName)
```
matt-crowson-moj, line 27: Can we not get rid of this class?  Don't think it is used anymore.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsImporter.kt
```diff
@@ -32,4 +32,6 @@ class LocationsImporter(private val locationRepo: LocationRepository, private va
             }
         }
     }
+
+    fun allImported() = locationRepo.findAll()
```
matt-crowson-moj, line 36: I am a bit unsure on this change and whether this should be here.  Perhaps alternatively return them via the import() method?

matt-crowson-moj (approved):
Approved but do have a question on the LocationImporter.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/location/importer/LocationsImporter.kt
jcavell: Got rid of it

matt-crowson-moj (approved):
LGTM

PR #49, opened 2020-10-07
P4-2214 changes to include the prices used for the calculations to an extra tab in the output spreadsheet.

matt-crowson-moj:
This change adds an additional tab to the outputted spreadsheet to include the prices from the suppliers JPC price book.  This help the assurance team review move prices and make adjustments by referring to the latest JPC price book


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/output/JPCPriceBookSheet.kt
```diff
@@ -0,0 +1,35 @@
...
+        this.forEach { sourceRow -> sourceRow.copyTo(destination.createRow(sourceRow.rowNum)) }
+    }
+
+    private infix fun Row.copyTo(destination: Row) {
+        this.forEach { sourceCell -> sourceCell.copyTo(destination.createCell(sourceCell.columnIndex)) }
+    }
+
+    private infix fun Cell.copyTo(destination: Cell) {
+        when (this.cellType) {
+            CellType.STRING -> destination.setCellValue(this.stringCellValue)
```
matt-crowson-moj, line 30: Not covering all possible fields types here.  This works fine for the current JPC price book.  We have test coverage so should anything change then we should know quickly.

jcavell (approved):

PR #68, opened 2020-10-19
Enable CI prod deployment

mafonso:
Enable CI to deploy to preprod and prod

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -8,21 +8,24 @@ image:
   tag: latest
   port: 8080
 
+db:
+  secret_name: rds-instance-calculate-journey-variable-payments-prod
+
 ingress:
   enabled: true
-  enable_whitelist: true
+  enable_whitelist: false
   hosts:
-    - host: calculate-journey-variable-payments.hmpps.service.justice.gov.uk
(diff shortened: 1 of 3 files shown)


helm_deploy/values-prod.yaml
```diff
@@ -8,21 +8,24 @@ image:
...
+  secret_name: rds-instance-calculate-journey-variable-payments-prod
+
 ingress:
   enabled: true
-  enable_whitelist: true
+  enable_whitelist: false
   hosts:
-    - host: calculate-journey-variable-payments.hmpps.service.justice.gov.uk
-      cert_secret: calculate-journey-variable-payments-cert
+    - host: calculate-journey-variable-payments-prod.apps.live-1.cloud-platform.service.justice.gov.uk
```
matt-crowson-moj, line 18: Not 100% sure on how the config works.  Do we need prod in the hostname?

matt-crowson-moj (approved):
LGTM

helm_deploy/values-prod.yaml
mafonso: We don't. It can be anything as long as it is unique.  I was just following the same logic as other projects.
The end goal it to have a custom domain, after I request the certificate.

PR #69, opened 2020-10-20
P4 2256 cancelled payable moves

jcavell:
Add cancelled moves to the spreadsheet.

The filterer ensures only cancelled moves that are cancelled after 3pm the day before the move are included. 
It creates a fake journey (because all pricing calculations expect a journey).

RowValue changed to add the move date, and nullable cancelled date and time.

Cancelled moves are added to a Cancelled sheet which is bespoke (it has different columns to the other tabs).
The summary tab includes cancelled moves now.

Localstack files added for cancelled moves so it can be tested using localstack.


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/output/CancelledMovesSheetTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import uk.gov.justice.digital.hmpps.pecs.jpc.config.JPCTemplateProvider
+import uk.gov.justice.digital.hmpps.pecs.jpc.location.LocationType
+import uk.gov.justice.digital.hmpps.pecs.jpc.pricing.Supplier
+import uk.gov.justice.digital.hmpps.pecs.jpc.reporting.*
+import java.time.LocalDate
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("test")
+@ContextConfiguration(classes = [TestConfig::class])
+internal class CancelledMovesSheetTest(@Autowired private val template: JPCTemplateProvider) {
```
matt-crowson-moj, line 22: You can reduce down to one  - `@SpringJUnitConfig(TestConfig::class)`

matt-crowson-moj (approved):
LGTM

PR #72, opened 2020-10-29
P4-2345, P4-2343, P4-2344, P4-2346, P4-2347

danielepolencic:
The following PR includes:

- The boilerplate code necessary to render the front-end pages
  - Thymeleaf, Layout dialect, prettytime.
- P4-2345
- P4-2343
- P4-2344
- P4-2346
- P4-2347

What it looks like:

image
image


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/constraint/MonthYearConstraintValidator.kt
```diff
@@ -0,0 +1,17 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.constraint
+
+import org.ocpsoft.prettytime.nlp.PrettyTimeParser
+import javax.validation.ConstraintValidator
+import javax.validation.ConstraintValidatorContext
+
+
+class MonthYearValidator : ConstraintValidator<ValidMonthYear?, String?> {
```
jcavell, line 8: I'm not sure what happens if the parse method bombs out. Just to be sure I've wrapped it in a runCatching:  
```
 override fun isValid(datesString: String?, context: ConstraintValidatorContext) =
            Result.runCatching { PrettyTimeParser().parse(datesString).size == 1}.getOrElse { false }
```

}

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtlmController.kt
```diff
@@ -0,0 +1,61 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.controller
+
+import org.joda.time.LocalDate
```
jcavell, line 3: Use java.time instead

jcavell (changes requested):
Could do with some controller tests, which I'm happy to look at after you make those small changes I've commented on.

jcavell (approved):

PR #76, opened 2020-11-03
P4-2434 adding ability to run an import on application start.

matt-crowson-moj:
The driver for this change is so we can run an import of locations and prices via direct call to jar file and negate the need to go via a rest call.  This should enable us to then call via a cron job for example.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/JpcApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/JpcApplication.kt
@@ -1,11 +1,20 @@
 package uk.gov.justice.digital.hmpps.pecs.jpc
 
+import org.springframework.beans.factory.getBean
+import org.springframework.boot.SpringApplication
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
+import uk.gov.justice.digital.hmpps.pecs.jpc.import.LocationAndPriceImporter
 
 @SpringBootApplication
 class JpcApplication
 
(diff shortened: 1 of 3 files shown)


mafonso:
I suppose this is something we want to build and have the .jar available in the (same) Docker container, so that we can call it with a different CMD/ENTRYPOINT. 

Maybe it makes sense to add commit for the Dockerfile and CI tests in the same PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/JpcApplication.kt
```diff
@@ -1,11 +1,20 @@
...
 
 @SpringBootApplication
 class JpcApplication
 
 fun main(args: Array<String>) {
-	runApplication<JpcApplication>(*args)
+    runApplication<JpcApplication>(*args).let { context ->
+
+        // This is a temporary solution to run an import of locations and prices then terminate bypassing the need to go to an endpoint.
+        args.firstOrNull { it == "--run-import" }?.let {
```
jcavell, line 16: We'll want to separate the import of locations/prices from the import of reports, so I'd suggest calling the parameter --import-locations-and-prices

jcavell (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/JpcApplication.kt
matt-crowson-moj: Done.

jcavell (approved):

PR #78, opened 2020-11-04
P4-2434 removed endpoints for importing locations and prices.  This can now only be done via the commandline.

matt-crowson-moj:
This changes means locations and prices can only be imported from the commandline going forwards.

@mafonso this change should mean going forwards we will not be restricted to one pod as this can be run via a cron job in a controlled manner.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/ImportController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/ImportController.kt
@@ -11,7 +11,6 @@ import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
 import uk.gov.justice.digital.hmpps.pecs.jpc.spreadsheet.SpreadsheetProtection
-import uk.gov.justice.digital.hmpps.pecs.jpc.price.Supplier
 import uk.gov.justice.digital.hmpps.pecs.jpc.service.ImportService
 import java.io.FileInputStream
 import java.io.IOException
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/import/LocationAndPriceImporterTest.kt
```diff
@@ -1,36 +1,60 @@
...
 import uk.gov.justice.digital.hmpps.pecs.jpc.location.LocationRepository
 import uk.gov.justice.digital.hmpps.pecs.jpc.price.PriceRepository
 import uk.gov.justice.digital.hmpps.pecs.jpc.price.Supplier
 import uk.gov.justice.digital.hmpps.pecs.jpc.service.ImportService
 
-internal class LocationAndPriceImporterTest {
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("test")
+@ContextConfiguration(classes = [TestConfig::class])
+internal class LocationAndPriceImporterTest(@Autowired val priceRepository: PriceRepository,
```
matt-crowson-moj, line 23: So I changed this test to actually remove and persist for real.  This in turn raised something an issue so glad I did.  The price importer was doing a delete all prices which is shouldn't be doing.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/ImportController.kt
```diff
@@ -25,25 +24,6 @@ class ImportController(private val importService: ImportService,
                        private val timeSource: TimeSource,
                        private val spreadsheetProtection: SpreadsheetProtection) {
 
-
-    @GetMapping("/import-locations")
```
matt-crowson-moj, line 29: These endpoints are now history!

jcavell (approved):

PR #79, opened 2020-11-04
Use a single default profile

mafonso:
* Use a single default profile (application.yml) mapping environment variables.
* Adds a local .env configuration set


matt-crowson-moj (approved):
LGTM - much cleaner IMO

PR #80, opened 2020-11-04
P4-2434 changes to support running the import of supplier reports from the commandline.

matt-crowson-moj:
This is part two of moving imports (of reports) to the command line.

The following commandline parameter has now been added, see example below:

`--import-supplier-reports=2020-10-01,2020-10-02`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/import/SupplierReportsImporter.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/import/SupplierReportsImporter.kt
@@ -3,18 +3,15 @@ package uk.gov.justice.digital.hmpps.pecs.jpc.import
 import org.slf4j.LoggerFactory
 import org.springframework.boot.ExitCodeGenerator
 import org.springframework.stereotype.Component
-import uk.gov.justice.digital.hmpps.pecs.jpc.location.LocationRepository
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/import/SupplierReportsImporter.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    private val failure = ExitCodeGenerator { 1 }
+
+    /**
+     * Calling this kicks off an import and returns '0' if successful or '1' if any exception is thrown (and caught).
+     */
+    fun import(from: LocalDate, to: LocalDate): ExitCodeGenerator {
+        return Result.runCatching {
+            importService.importReports(Supplier.GEOAMEY, from, to)
+            importService.importReports(Supplier.SERCO, from, to)
```
jcavell, line 28: Hmm, this makes me think I should change the import to import both suppliers rather than one at a time.

jcavell (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/import/SupplierReportsImporter.kt
matt-crowson-moj: Yeah so I wasn't sure on this.  I did originally think about doing each supplier separately...

PR #81, opened 2020-11-05
P4-2402 Choose a supplier

danielepolencic:
TL;DR: Select a supplier if you haven't done so before.

Screenshots:

image
image

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
@@ -6,10 +6,7 @@ import org.springframework.format.annotation.DateTimeFormat
 import org.springframework.stereotype.Controller
 import org.springframework.ui.ModelMap
 import org.springframework.validation.BindingResult
-import org.springframework.web.bind.annotation.ModelAttribute
-import org.springframework.web.bind.annotation.RequestMapping
-import org.springframework.web.bind.annotation.RequestMethod
-import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.*
 import org.springframework.web.servlet.view.RedirectView
 import uk.gov.justice.digital.hmpps.pecs.jpc.constraint.ValidMonthYear
 import uk.gov.justice.digital.hmpps.pecs.jpc.price.Supplier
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
```diff
@@ -31,11 +28,30 @@ class HtmlController(@Autowired val dashboardService: DashboardService) {
         return RedirectView("/dashboard")
```
jcavell, line 28: Should now redirect to choose-supplier?

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
```diff
@@ -31,11 +28,30 @@ class HtmlController(@Autowired val dashboardService: DashboardService) {
...
+    @RequestMapping("/choose-supplier/geoamey")
+    fun chooseSupplierGeoAmey(session: HttpSession): RedirectView {
+        session.setAttribute("supplier", Supplier.GEOAMEY)
+        return RedirectView("/dashboard")
+    }
+
     @RequestMapping("/dashboard")
-    fun dashboard(@RequestParam(name = "date", required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) date: LocalDate?, model: ModelMap): String {
+    fun dashboard(@SessionAttribute(name = "date") @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) date: LocalDate?, @SessionAttribute(name = "supplier") supplier: Supplier?, model: ModelMap): Any {
+        if (supplier == null) {
```
jcavell, line 50: Because this is going to be a repeated pattern per RequestMapping (apart from choose-supplier), probably better to put the check for supplier in a Filter - see https://stackoverflow.com/questions/57137095/spring-mvc-session-check-before-every-request

jcavell (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
jcavell: Maybe not

jcavell (approved):

PR #82, opened 2020-11-05
P4 2368 summary page better queries

jcavell:
Instead of having a single query that returns all moves for the month, and then calculating unique journeys and summaries from this in the code, leverage separate (somewhat complex) queries for the summary and the unique journeys.

This makes the summary page quicker and doesn't result in all moves being loaded in for the month to do the calculation.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/spreadsheet/PricesSpreadsheetGenerator.kt
```diff
@@ -12,61 +12,64 @@ import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
...
 
-            val moves = moveQueryRepository.findSummaryForSupplierInDateRange(filter.supplier, filter.movesFrom, filter.movesTo)
+            val moves = dashboardService.movesForMonth(supplier, startDate)
+            val summaries = dashboardService.summariesForMonth(supplier, startDate)
 
-            with(moves) {
                 StandardMovesSheet(workbook, header)
                         .also { logger.info("Adding standard prices.") }
-                        .apply { writeMoves(standard) }
+                        .apply { writeMoves(moves[0]) }
```
matt-crowson-moj, line 43: Curious about use of index....how do you know it is the correct one?

matt-crowson-moj (dismissed):
LGTM

matt-crowson-moj (approved):
LGTM

PR #83, opened 2020-11-05
P4-2434 removal of import supplier reports endpoint as to be run via commandline.

matt-crowson-moj:
This changes means you you can only import supplier reports from the command line going forwards, for example:

`java -jar app.jar --spring.main.web-application-type=none --import-supplier-reports=2020-10-01,2020-10-02`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
@@ -1,13 +1,13 @@
 package uk.gov.justice.digital.hmpps.pecs.jpc.config
 
-import org.springframework.context.annotation.Profile
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
```diff
@@ -1,13 +1,13 @@
...
-import org.springframework.context.annotation.Profile
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication
 import org.springframework.security.config.annotation.web.builders.HttpSecurity
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter
 import org.springframework.security.config.web.servlet.invoke
 
-// TODO for now this is tied to run for localstack profile only!!
 @EnableWebSecurity
+@ConditionalOnWebApplication
```
matt-crowson-moj, line 10: @jcavell This means we can avoid starting web server.

jcavell (approved):

PR #91, opened 2020-11-09
P4-2434 bringing the test package structure in-line with the implementations.

matt-crowson-moj:
@jcavell this is really just a refactoring PR.  The tests were not in-line with the src.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/JpcApplication.kt
```diff
@@ -4,8 +4,8 @@ import org.springframework.beans.factory.getBean
 import org.springframework.boot.SpringApplication
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
-import uk.gov.justice.digital.hmpps.pecs.jpc.import.LocationAndPriceImporter
-import uk.gov.justice.digital.hmpps.pecs.jpc.import.SupplierReportsImporter
+import uk.gov.justice.digital.hmpps.pecs.jpc.importer.LocationAndPriceImporter
```
matt-crowson-moj, line 7: A note on this package name.  `import` is technically a reserved word and was causing issues when trying to refactor so renamed to `importer`.

jcavell (approved):

PR #93, opened 2020-11-09
P4-2434 adding date default for supplier reports when running via a cron.

matt-crowson-moj:
Tweaked the command line option for running supplier reports to default the dates if none supplied.  Will default to a days worth of data on the previous day of the current date (the date when run).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,20 @@ export $(cat .env | xargs)  # If you want to set or update the current shell env
 java -jar app.jar --spring.main.web-application-type=none --import-locations-and-prices
 ```
 
-Importing supplier reports for given dates:
+
+Importing supplier reports with no supplied dates:
+
+*Note: using this defaults to one days worth of report data ending on the previous day to today's date.  If the current date is 30th November then from and to would be from 28th to 29th November.*
+
+```bash
+export $(cat .env | xargs)  # If you want to set or update the current shell environment
+
+java -jar app.jar --spring.main.web-application-type=none --import-supplier-reports
+```
+
+Importing supplier reports with supplied dates:
+
+*Note: care should be taken when supplying the date range.  This should not go beyond a month period due to the volume of reporting data*
 
 ```bash
 export $(cat .env | xargs)  # If you want to set or update the current shell environment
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/JpcApplication.kt
```diff
@@ -19,12 +20,23 @@ fun main(args: Array<String>) {
...
+                val to = LocalDate.parse(dates.split(",")[1].trim())
 
-            (context.getBean(SupplierReportsImporter::class) as SupplierReportsImporter).let { reportImporter ->
-                SpringApplication.exit(context, reportImporter.import(from, to)) }
+                (context.getBean(SupplierReportsImporter::class) as SupplierReportsImporter).let { reportImporter ->
+                    SpringApplication.exit(context, reportImporter.import(from, to)) }
+            } else {
+                (context.getBean(TimeSource::class) as TimeSource).let { ts ->
+                    val to = ts.date().minusDays(1)
+                    val from = to.minusDays(1)
```
jcavell, line 33: A bit confused as to why you want to default this, and to just one day - is that for testing or for some other reason?

README.md
```diff
@@ -49,7 +49,20 @@ export $(cat .env | xargs)  # If you want to set or update the current shell env
...
+
+```bash
+export $(cat .env | xargs)  # If you want to set or update the current shell environment
+
+java -jar app.jar --spring.main.web-application-type=none --import-supplier-reports
+```
+
+Importing supplier reports with supplied dates:
+
+*Note: care should be taken when supplying the date range.  This should not go beyond a month period due to the volume of reporting data*
```
jcavell, line 65: At the moment it unfortunately will need to be set for longer than a month period, because of the backfill issue. Data for a move can currently span months, which is a problem. That's one of the reasons I'm thinking we'll need to do incremental imports, persisting the reporting data for each move.

matt-crowson-moj:
@mafonso chatted with Jonny.  We are not going to do this right now.  So forget the cron.

PR #99, opened 2020-11-12
P4-2452 initial work to enforce role access to the application.

matt-crowson-moj:
So the work done for this change is not really ideal.  Struggled to get Spring to pick up the token authorities.  I think it works for now but it may need to be revisited.

A fake user has also been added to the[ hmpps-auth service](https://github.com/ministryofjustice/hmpps-auth/blob/main/src/main/resources/db/dev/data/auth/V900_0__oauth_sample_data.sql#L255), **JPC_USER**.  So to run the app on a dev machine you will need the latest hmpps-auth container running on port 9090.


jcavell (approved):
As discussed, it would be good to add the jwt as an attribute to DefaultOAuth2User, then it ends up as an attribute on SPRING_SECURITY_CONTEXT.. You could then create a JwtFilter to look at this value for each request and ensure it hasn't expired.

PR #104, opened 2020-11-19
P4-2238 - Write journeys tab on output spreadsheet

jcavell:
JPC_template.xlsx updated to contain the new tab

JourneysSheet - generates the rows for the new journey tab
Called from PricesSpreadsheetGenerator, displays the unique journeys
PricesSpreadsheetGenerator gets journeys from MoveService

MoveQueryRepository
uniqueJourneysInDateRange takes excludePriced parameter (defaults to true)

UniqueJourney
Add methods for unitPrice in pence / pounds and from / to site name

Person
Change latestNomisBookingId type to be Int? not String?


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/spreadsheet/PricesSpreadsheetGenerator.kt
```diff
@@ -60,6 +61,10 @@ class PricesSpreadsheetGenerator(@Autowired private val template: JPCTemplatePro
                         .also { logger.info("Adding cancelled moves.") }
                         .apply { writeMoves(moves[5]) }
 
+            JourneysSheet(workbook, header)
+                    .also { logger.info("Adding journeys.") }
+                    .apply { writeJourneys(journeys) }
```
matt-crowson-moj, line 66: Could inline call to get journeys I suppose so only hold in memory/scope as and when needed?

matt-crowson-moj (approved):
LGTM

PR #106, opened 2020-11-23
P4 2435 improve import

jcavell:
Allows import to be run one day at a time, so that you can run:

`java -jar ./build/libs/calculate-journey-variable-payments-2020-11-22.jar --import-supplier-reports=2020-10-02,2020-10-02`

and then the next day
`java -jar ./build/libs/calculate-journey-variable-payments-2020-11-22.jar --import-supplier-reports=2020-10-03,2020-10-03`

It merges new updates, and journeys and events into the previously persisted entities.

Also, runs a single import for both suppliers now.

You are still able to run the import for any date range, but note that it is expected that the most recently run import is the latest (i.e. don't go back and run an import for a day in Sept and expect this to be a good thing).

Moves (and their journeys and events) are now stored in the database even if they aren't completed / cancelled. In other words, ALL moves are stored because there may be be an event or update in the future that puts it into a completed state ready for the JPC. Therefore, a lot more moves are now stored (nearly 100k on prod).

The mapping of the the Move / Journey into a ReportMove and ReportJourney and back again is a little bit painful, so in the future it might be good to look at consolidating this into a single entity, like for Event.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
```diff
@@ -4,94 +4,124 @@ import org.slf4j.LoggerFactory
...
-                        )
-
-                        // add journeys to move model
-                        val journeyModels = report.journeysWithEvents.map { journeyModel(moveModel.moveId, it) }
-                        moveModel.addJourneys(*journeyModels.toTypedArray())
-
-                        moveRepository.save(moveModel)
-
-                    }.onFailure { logger.warn(it.message) }
+    fun persist(reports: List<Report>) {
```
matt-crowson-moj, line 13: A bit on the large size.  Would be nice to just break this up a little...

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/ImportService.kt
```diff
@@ -18,22 +17,20 @@ class ImportService(
...
     private val logger = LoggerFactory.getLogger(javaClass)
 
     fun importLocations() = import(locationsImporter::import)
 
     fun importPrices(supplier: Supplier) = import { priceImporter.import(supplier) }
 
-    fun importReports(supplier: Supplier, reportsFrom: LocalDate, reportsTo: LocalDate) {
-        logger.info("Importing reports for supplier '$supplier', moves from '$reportsFrom', moves to '$reportsTo'.")
+    fun importReports(reportsFrom: LocalDate, reportsTo: LocalDate) {
+        logger.info("Importing reports from '$reportsFrom', moves to '$reportsTo'.")
```
matt-crowson-moj, line 29: Minor comment.  Should this say **reports to** and not **moves to**?

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/ImportService.kt
```diff
@@ -18,22 +17,20 @@ class ImportService(
...
 
-        val reports = import { reportImporter.import(supplier, reportsFrom, reportsTo) }
+        val reports = import { reportImporter.import(reportsFrom, reportsTo) }
 
-        reports?.let {
-            reportModelPersister.persist(FilterParams(supplier, reportsFrom, reportsTo), it.toList())
-        }
+        reports?.let { reportPersister.persist(it.toList()) }
```
matt-crowson-moj, line 33: Might useful just to log if not importing anything...

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersisterTest.kt
```diff
@@ -41,83 +32,63 @@ internal class MovePersisterTest {
...
-
-    @Autowired
-    lateinit var journeyRepository: JourneyRepository
-
     @Autowired
     lateinit var entityManager: TestEntityManager
 
-    val from = LocalDate.of(2020, 9, 1)
-    val to = LocalDate.of(2020, 9, 6)
+    final val from: LocalDate = LocalDate.of(2020, 9, 1)
```
matt-crowson-moj, line 38: Why **final**???  It's a **val**, am I missing something?

matt-crowson-moj (approved):
LGTM

PR #107, opened 2020-11-23
P4-2452 flushing data after each repository call in effort to reduce memory footprint.

matt-crowson-moj:
Not able to prove this locally.  Need to deploy to see if this resolves an out of memory issue when importing locations and prices spreadsheets.  Will revert if it doesn't.


mafonso (approved):
LGTM

PR #109, opened 2020-11-23
P4-2452 splitting out the imports further in effort to stop out of memory errors.

matt-crowson-moj:
As it stands keeping the import of locations and prices in one call is causing out of memory errors when running on the container.  This splits out each import into a separate call to try and avoid this.


mafonso (approved):
LGTM

PR #113, opened 2020-11-25
P4-2493 initial commit for start of work for mapping friendly location names.

matt-crowson-moj:
This is a draft PR to get some feed back.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/JourneyMapLocationController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/JourneyMapLocationController.kt
@@ -1,0 +1,45 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.controller
+
+import org.springframework.stereotype.Controller
+import org.springframework.ui.ModelMap
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.ModelAttribute
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.SessionAttributes
+import uk.gov.justice.digital.hmpps.pecs.jpc.price.Supplier
+import uk.gov.justice.digital.hmpps.pecs.jpc.service.JourneyMapLocationService
+
+@Controller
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/JourneyMapLocationController.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.SessionAttributes
+import uk.gov.justice.digital.hmpps.pecs.jpc.price.Supplier
+import uk.gov.justice.digital.hmpps.pecs.jpc.service.JourneyMapLocationService
+
+@Controller
+@SessionAttributes("supplier")
+class JourneyMapLocationController(private val service: JourneyMapLocationService) {
+
+  @GetMapping("/map-pickup-location/{journey-id}")
```
matt-crowson-moj, line 16: So I thought we would drive the mapping here by the journey id.  Does this makes sense in this context.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/JourneyMapLocationController.kt
```diff
@@ -0,0 +1,45 @@
...
+  fun mapPickupLocation(@PathVariable("journey-id") journeyId: String,
+                        @ModelAttribute(name = "supplier") supplier: Supplier,
+                        model: ModelMap): String {
+
+    val agencies = service.findPickUpAndDropOffAgenciesForJourney(journeyId)
```
matt-crowson-moj, line 21: So the other question I have is around how do we decided (if at all) whether mapping is allowed i.e. the nomis agency ids already have friendly names?

src/main/resources/templates/map-location.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
```
matt-crowson-moj, line 1: Will need a bit of TLC.  Hoping Daniele can help here.

PR #120, opened 2020-11-30
p4-2535 fixed missed env config in the helm _env.tpl file.

matt-crowson-moj:


mafonso (approved):
LGTM
Checked the config is present in all envs.

PR #122, opened 2020-11-30
P4-2535 correcting the sql script for the shedlock table.

matt-crowson-moj:
Using incorrect script for the shedlock table.  Needs a primary key.


mafonso (approved):
LGTM

PR #123, opened 2020-11-30
P4-2535 config cleanup for flyway.

matt-crowson-moj:
Not 100% sure this fixes the dev deployment.  It does seem however that removing this config does not break the local build.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -41,12 +41,6 @@ spring:
 
   flyway:
     locations: classpath:/db/migration/postgres,classpath:/db/migration
-    url: ${APP_DB_URL}
-    user: ${DB_USERNAME}
-    password: ${DB_PASSWORD}
-    placeholders:
-      database_password: ${DB_PASSWORD}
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yml
```diff
@@ -41,12 +41,6 @@ spring:
 
   flyway:
     locations: classpath:/db/migration/postgres,classpath:/db/migration
-    url: ${APP_DB_URL}
-    user: ${DB_USERNAME}
-    password: ${DB_PASSWORD}
-    placeholders:
-      database_password: ${DB_PASSWORD}
-    schemas: ${DB_NAME}
```
mafonso, line 49: Does it not need any?  will it use the existing database config?

matt-crowson-moj: Running locally it did.  I am still not 100% it is quite right though.  So Spring (automagically) integrates with flyway i.e. it should pick these up as needed (in theory).

danielepolencic (approved):

mafonso (approved):
LGTM

PR #128, opened 2020-12-01
P4-2441 adding some extra logging to see if we have a performance issue with the reports importer.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
```diff
@@ -75,11 +77,17 @@ class MovePersister(private val moveRepository: MoveRepository) {
 
                     moveRepository.save(newMove)
 
+                    if (counter % 500 == 0) {
+                        logger.info("Persisted $counter moves.  Flushing moves to the database.")
+
+                        moveRepository.flush()
```
matt-crowson-moj, line 83: @jcavell this might help reduce the memory footprint...

jcavell (approved):

PR #130, opened 2020-12-02
P4-2535 fixing the date for the reporting data scheduled import.

matt-crowson-moj:
To import a singles day reports then the from and to dates should be the same.  The changes fixes that on the cron job.


mafonso (approved):
LGTM

PR #149, opened 2020-12-10
Fix Journey Search

jcavell:
Manage JPC should show all pricing, including for journeys that haven't been made.

This commit addresses that by removing the journeys table from the query.

Also, Manage JPC now takes effective year from the currently selected year.


matt-crowson-moj (approved):
LGTM

PR #151, opened 2020-12-11
P4-2551 Fix Spreadsheet issue

jcavell:
Fix the amount of time it takes to download the spreadsheet.
Do not create CellStyles at run-time; instead create the ones we need once
Create CellStyle once for each color / format combination.

Remove limit from queries so all moves are shown

Remove developer notes from Excel Template

Ensure template is saved on summary view


matt-crowson-moj (approved):
LGTM

PR #159, opened 2020-12-16
P4-2551 spreadsheet changes.  Background row colour now solid, changed text on journeys tab. If no unit price on journeys tab then display 'NOT PRESENT'

matt-crowson-moj:
Mainly output pricing spreadsheet changes (and config):

- Solid fill colour used for alternate rows.
- NOT PRESENT display on Journeys tab unit price column if null or 0.
- Updated header text on Journeys tab from 'TOTAL VOLUME BY LOCATION' to 'TOTAL VOLUME BY JOURNEY'
-  Changed the DDL creation from 'create' to 'update'.


src/main/resources/application.yml
```diff
@@ -5,7 +5,7 @@ spring:
     database: POSTGRESQL
     show-sql: false
     hibernate:
-      ddl-auto: create
+      ddl-auto: update
```
matt-crowson-moj, line 8: We need to change this back so as not to delete everything.

jcavell (approved):

PR #160, opened 2020-12-16
P4-22533 Improve import

jcavell:
**ddl-auto set to create - will re-create all database tables**

Improve import speed: add batch select of moves and associated journeys and events in chunks of 100
----
Unsuccessful (but non-harmful) attempt to batch inserts / updates by adding the following to application.yml:
    properties:
      hibernate.jdbc.batch_size: 5
      hibernate.order_inserts: true
      hibernate.order_updates: true
      hibernate.jdbc.batch_versioned_data: true
      hibernate.generate_statistics: false

In conjunction with applying saveAll to a collection of entities.

MovePersister successfully improved to batch select moves and associated journeys and events in chunks of 100

Add move_id index to Journey
Add eventable_id index to Event

Fix bug in move.html where the Journey Event JSON was being displayed if there was only a single Journey Event

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
@@ -20,42 +20,57 @@ class MovePersister(private val moveRepository: MoveRepository,
     fun persist(moves: List<Move>) {
 
         var counter = 1
+        val movesToSave = mutableListOf<Move>()
+        val journeysToSave = mutableListOf<Journey>()
+        val eventsToSave = mutableListOf<Event>()
         logger.info("Persisting ${moves.size} moves")
 
-        moves.forEach { newMove ->
+        fun save() {
+            saveFlushAndClear(moveRepository, movesToSave)
+            saveFlushAndClear(journeyRepository, journeysToSave)
+            saveFlushAndClear(eventRepository, eventsToSave)
+        }
+
+        moves.chunked(100).forEach { hundredMoves ->
+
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
```diff
@@ -106,19 +118,19 @@ class MovePersister(private val moveRepository: MoveRepository,
 
     fun fakeCancelledJourney(move: Move): Journey {
```
matt-crowson-moj, line 119: Would be handy to document why we need this method (fake journey).

matt-crowson-moj, line 119: Would be handy to document why we need this method (fake journey).

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/Persister.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.move
+
+import org.springframework.data.jpa.repository.JpaRepository
+
+fun <T>saveFlushAndClear(repo: JpaRepository<T, String>, entities: MutableCollection<T>){
+    repo.saveAll(entities)
+    repo.flush()
+    entities.clear()
```
matt-crowson-moj, line 8: Does this `clear` make a difference?

matt-crowson-moj (approved):
LGTM

PR #165, opened 2020-12-17
P4-2529 Search by move reference ID

jcavell:
**This adds a link to "search moves by move reference ID" on the move summary page**

New findByReference method in MoveRepository
Used by findMoveByReference in MoveService

HtmlController has new findMove GET and performFindMove POST methods
The reference is validated. Unsuccessful search redirects back to the find a move search form with appropriate message.

Successful search redirects to the moves details page for the found move_id UUID

Queries for GET move and find move now restrict by supplier.

Add 404 page to return if a move doesn't exist (for that supplier)


matt-crowson-moj (approved):
LGTM

PR #170, opened 2021-01-07
Update README to include EnvFile setup

Thource:
Change psql port to 5433 to avoid clashes with local installs
Add AWS vars to .env


matt-crowson-moj (approved):
LGTM

PR #187, opened 2021-01-20
P4-2613 resolving the ordering of journey events. Was using the order they were pulled from the DB.

matt-crowson-moj:
Changes:

- Order the the journey events based on their pickup time for the presentation layer.  Was originally just using the order in which they were pulled from the DB hence sometimes it worked and other times it didn't.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/MoveService.kt
```diff
@@ -26,7 +26,7 @@ class MoveService(
         eventRepository.findByEventableIdIn(it.journeys.map { j -> j.journeyId }).groupBy { e -> e.eventableId }
 
       val journeysWithEvents = it.journeys.map { journey ->
-        journey.copy(events = journeyId2Events[journey.journeyId] ?: listOf())
+        journey.copy(events = journeyId2Events[journey.journeyId]?.sortedBy { e -> e.occurredAt } ?: listOf())
```
matt-crowson-moj, line 29: This is the key bit sorting the journey events.

Thource (approved):

PR #191, opened 2021-01-26
P4-2518 changes to clean up session handling e.g. timeouts, max concurrent sessions etc.

matt-crowson-moj:
Changes:

- Reduced session inactivity timeout to 30 mins.
- Max concurrent sessions for a user limited to one.
- Sessions are now automatically invalidated for any logins that fail.  Instead of waiting for the Spring background cleanup process.
- Session times now redirect back to the auth service.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
@@ -31,7 +31,7 @@ class SecurityConfiguration : WebSecurityConfigurerAdapter() {
   private lateinit var issuer: String
 
   @Value("\${HMPPS_AUTH_BASE_URI}")
-  private lateinit var authLogoutSuccesUri: String
+  private lateinit var authLogoutSuccessUri: String
 
   @Throws(Exception::class)
   override fun configure(http: HttpSecurity) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
```diff
@@ -64,19 +72,25 @@ class SecurityConfiguration : WebSecurityConfigurerAdapter() {
...
       val auth: Authentication? = SecurityContextHolder.getContext().authentication
 
-      if (auth != null) {
-        logger.warn("User: ${auth.name} attempted to access the protected URL: ${request.requestURI}")
-      }
+      auth?.let { logger.warn("User: ${auth.name} attempted to access the protected URL: ${request.requestURI}") }
 
       logger.error(accessDeniedException.message)
+
+      request.session.invalidate()
```
matt-crowson-moj, line 91: Without this an unused session will sit around waiting for it to he cleaned up.  There is no need as this can be cleaned up straight away.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
```diff
@@ -41,16 +41,24 @@ class SecurityConfiguration : WebSecurityConfigurerAdapter() {
         authorize("/info", permitAll)
         authorize(anyRequest, hasRole("PECS_JPC"))
       }
+      sessionManagement {
+        invalidSessionUrl = authLogoutSuccessUri
+        sessionAuthenticationErrorUrl = authLogoutSuccessUri
+        sessionConcurrency {
+          maximumSessions = 1
```
matt-crowson-moj, line 48: Limits maximum number of concurrent session to one for a given user.  The knock on effect of this is if a user logs in a second time then then their other session will be automatically ended.

PR #195, opened 2021-01-28
P4-2518 correction after issue with log level was spotted release notes in the DPS gradle plugin.

matt-crowson-moj:
Changes:

- The original file where this was taken from in the dps gradle spring plugin had the wrong log level.  This change corrects it.  In truth is has little impact at this point in time as we are not using Application Insights at time of writing.

### DIFF
--- a/applicationinsights.json
+++ b/applicationinsights.json
@@ -1,12 +1,16 @@
 {
   "role": {
-    "name": "application-name"
+    "name": "calculate-journey-variable-payments"
   },
   "customDimensions": {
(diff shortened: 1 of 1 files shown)


applicationinsights.json
```diff
@@ -1,12 +1,16 @@
 {
   "role": {
-    "name": "application-name"
+    "name": "calculate-journey-variable-payments"
```
matt-crowson-moj, line 3: The application name was an omission on my part in the original commit of this file.

Thource (approved):

PR #201, opened 2021-02-05
P4-2654 updating move query to exclude moves without a move type.  These are incomplete (unpriced) moves in JPC.

matt-crowson-moj:
Changes:

- Updated query to retrieve moves for viewing in the application to only fetch complete and/or cancelled moves i.e. moves which have a move_type and are not null.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MoveQueryRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MoveQueryRepository.kt
@@ -176,9 +176,12 @@ class MoveQueryRepository(@Autowired val jdbcTemplate: JdbcTemplate) {
     }
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MoveQueryRepository.kt
```diff
@@ -176,9 +176,12 @@ class MoveQueryRepository(@Autowired val jdbcTemplate: JdbcTemplate) {
     }
   }
 
+  /**
+   * Excludes moves without a move_type. This is an indication of bad reporting data in the JSON data feeds.
```
matt-crowson-moj, line 180: Strictly speaking we shouldn't be sent bad data but this the situation we are currently in.  It was a conscious decision to keep accepting these, at least in the short term, as it is an indicator of being supplied bad/poor/incomplete move data.

Thource (approved):

PR #206, opened 2021-02-11
P4-2635 resolves issues for standard moves where locations should match, if not then flag as multi-type move.

matt-crowson-moj:
Changes:

- The point at which moves are identified for persistence during import (for standard moves) has been updated to check the journey locations match that of the move.  If they don't then the move should fall under the multi-type move category.

**_Note: When this change is merged it is likely we will need to re-import backdated reporting data so this change can be applied to historic moves (i.e. update them where applicable)._**


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/report/MoveFilterer.kt
```diff
@@ -29,28 +30,32 @@ object MoveFilterer {
...
    */
-  fun isStandardMove(move: Move): Boolean {
-    return isCompleted(move) &&
-      with(move.journeys.map { it }) {
+  fun isStandardMove(move: Move) =
+    isCompleted(move) &&
+      with(move.journeys) {
         count { it.stateIsAnyOf(JourneyState.completed) } == 1 &&
-          count { it.stateIsAnyOf(JourneyState.completed) && it.billable } == 1 &&
+          count { isCompleteBillableJourneyAndLocationsMatchMove(it, move) } == 1 &&
```
matt-crowson-moj, line 44: This is and the prive fun below are the key changes for this PR.  The other bits are some small tidy-ups/refactorings.

Thource (dismissed):

Thource (approved):

PR #208, opened 2021-02-16
P4-2666 changes to (fix) improve the robustness of the move import process.

matt-crowson-moj:
Changes:

- Increase lengths of certain DB columns that caused the failure of the importer due to being too long (using DB migration).
- Reworked the MovePersister to not fail and exit if an move fails to import.  It will skip any bad moves.
- Move the MovePersister away from batch importing of moves.  Will now push a single move to the DB with is relevant other data.
- Reduced the batch size of the Person/Profile importer (Will actually look at moving away from batch inserts in a separate PR).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/report/Event.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/report/Event.kt
@@ -59,12 +59,9 @@ data class Event constructor(
   @Column(name = "recorded_at")
   val recordedAt: LocalDateTime,
 
-  var notes: String?,
+  @Column(nullable = true, length = 1024)
+  val notes: String?,
 ) : Comparable<Event> {
-
-  init {
-    notes = notes?.take(255)
-  }
 
   fun hasType(et: EventType) = type == et.value
   override fun equals(other: Any?): Boolean {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.pecs.jpc.move
 
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
 import uk.gov.justice.digital.hmpps.pecs.jpc.importer.report.Event
 import uk.gov.justice.digital.hmpps.pecs.jpc.importer.report.EventType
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/report/Event.kt
```diff
@@ -59,13 +59,10 @@ data class Event constructor(
   @Column(name = "recorded_at")
   val recordedAt: LocalDateTime,
 
-  var notes: String?,
+  @Column(name = "metadata", nullable = true, length = 1024)
```
matt-crowson-moj, line 62: Default was 255.  This was too short.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/report/Event.kt
```diff
@@ -59,13 +59,10 @@ data class Event constructor(
...
+  val notes: String?,
 ) : Comparable<Event> {
 
-  init {
```
matt-crowson-moj, line 65: Scraped this.  The JSON parser was ignoring this anyway.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
```diff
@@ -21,12 +22,13 @@ class MovePersister(
...
   fun persist(moves: List<Move>) {
 
-    var counter = 1
+    var counter = 0
     val movesToSave = mutableListOf<Move>()
     val journeysToSave = mutableListOf<Journey>()
     val eventsToSave = mutableListOf<Event>()
     logger.info("Persisting ${moves.size} moves")
 
+    @Transactional
```
matt-crowson-moj, line 31: Moved the transaction boundary to be at the single move level.  This was needed as we don't want to batch insert moves.  The overhead is slower as more network trips.  The big plus is robustness.  It will only fail on the specific move and not the whole import process.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
```diff
@@ -90,14 +92,26 @@ class MovePersister(
 
           eventsToSave += moveEvents + journeyEvents
 
-          if (counter++ % 1000 == 0) {
-            logger.info("Persisted $counter moves out of ${moves.size} (flushing moves to the database).")
-            save()
+          Result.runCatching { save() }
```
matt-crowson-moj, line 95: Another key change here is to not fail if the save actually fails.  We want to continue.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MovePersister.kt
```diff
@@ -90,14 +92,26 @@ class MovePersister(
...
+            .onSuccess { counter++ }
+            .onFailure {
```
matt-crowson-moj, line 97: Undecided on how to report the failed move at the moment so just logging for now.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/PersonPersister.kt
```diff
@@ -19,7 +19,7 @@ class PersonPersister(
     val peopleToSave = mutableListOf<Person>()
     people.forEach { person ->
       peopleToSave += person
-      if (++counter % 500 == 0) {
+      if (++counter % 50 == 0) {
```
matt-crowson-moj, line 22: I am actually going to change this to work in a similar way to the move persister.  This still lacks robustness.

Thource (approved):

PR #216, opened 2021-02-22
P4-2657 some house keeping bumping to later 3rd party libraries and adding some logging around journey and price retrieval.

matt-crowson-moj:
Changes:

- Non-functional changes.  Bumping 3rd party libraries.
- Added logging where logger was specified but not used in the JourneyService.  This will aid us with some basic retrieval metrics.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/JourneyService.kt
```diff
@@ -11,20 +13,46 @@ class JourneyService(private val journeyQueryRepository: JourneyQueryRepository)
 
   private val logger = LoggerFactory.getLogger(javaClass)
```
matt-crowson-moj, line 14: This was specified but not actually used (it is now).

Thource (approved):

PR #219, opened 2021-02-23
Add effectiveYear to delete criteria on import-prices

Thource:


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/price/PriceImporter.kt
```diff
@@ -20,18 +22,18 @@ class PriceImporter(
 
   private val logger = LoggerFactory.getLogger(javaClass)
 
-  fun import(supplier: Supplier) {
+  fun import(supplier: Supplier, timeSource: TimeSource) {
```
matt-crowson-moj, line 25: Sorry, TimeSource should be wired in via the constructor, it is only done once then and we don't need to keep passing it in, so....

`class PriceImporter(private val timeSource: TimeSource, .....)`

matt-crowson-moj (changes requested):
All fine a part from the constructor comment which will have slight impact on the others classes touched now.

matt-crowson-moj (approved):
LGTM

PR #223, opened 2021-02-26
P4 2674/build in resilience for failed person profile import

matt-crowson-moj:
Changes:

- Removal of batch inserts for people and profiles and stopping the import process from failing should an insert into the database fail.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/ImportService.kt
```diff
@@ -14,7 +14,6 @@ import java.time.Duration
 import java.time.LocalDate
 
 @Service
-@Transactional
```
matt-crowson-moj, line 17: So I put this back in by mistake.  It needs to be at a lower level.

PR #226, opened 2021-03-04
P4-2675 changes to fix an issue with the price importer throwing a duplicate price exception.

matt-crowson-moj:
Changes:

- This changes the price importer to check existing prices effective year.  Without this we can get duplicate errors when really there shouldn't be.  For example we can have a 2019 price in the system where the **from** and **to** are the same as a 2020 price we are trying to import, this is not a duplicate!

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/price/PricesSpreadsheet.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/price/PricesSpreadsheet.kt
@@ -22,7 +22,8 @@ class PricesSpreadsheet(
   private val spreadsheet: Workbook,
   val supplier: Supplier,
   supplierLocations: List<Location>,
-  private val pricesRepository: PriceRepository
+  private val pricesRepository: PriceRepository,
+  private val effectiveYear: Int
 ) : InboundSpreadsheet(spreadsheet) {
 
   val errors: MutableList<PricesSpreadsheetError> = mutableListOf()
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/price/PricesSpreadsheet.kt
```diff
@@ -58,20 +59,23 @@ class PricesSpreadsheet(
...
       fromLocation = fromLocation,
       toLocation = toLocation,
       priceInPence = price,
-      effectiveYear = 2020 // TODO remove this once we're not importing from spreadsheet anymore
+      effectiveYear = effectiveYear
     )
   }
 
+  private fun failIfPriceAlreadyExists(supplier: Supplier, fromLocation: Location, toLocation: Location) {
+    pricesRepository.findBySupplierAndFromLocationAndToLocationAndEffectiveYear(supplier, fromLocation, toLocation, effectiveYear)?.let {
```
matt-crowson-moj, line 74: This is the key change here.  We need to check the new price against existing prices including the effective year.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/price/PricesSpreadsheet.kt
```diff
@@ -58,20 +59,23 @@ class PricesSpreadsheet(
...
-      throw RuntimeException("Duplicate price: '${fromLocation.siteName}' to '${toLocation.siteName}' for $supplier")
-    }
+    failIfPriceAlreadyExists(supplier, fromLocation, toLocation)
 
     return Price(
       supplier = supplier,
       fromLocation = fromLocation,
       toLocation = toLocation,
       priceInPence = price,
-      effectiveYear = 2020 // TODO remove this once we're not importing from spreadsheet anymore
```
matt-crowson-moj, line 71: The reason this was here is down to the fact that price importing manually was supposed to be short lived functionality to help the end users i.e. no longer applicable.  However uptake/use of the system has been slow.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/price/PricesSpreadsheet.kt
Nimphal: So is the effective year a column in the spreadsheet? Sorry, very out of the loop here.

Nimphal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/importer/price/PricesSpreadsheet.kt
matt-crowson-moj: Hopefully my ramblings on Slack have helped here....

PR #230, opened 2021-03-09
P4-2691 resolve issue where by a price for the same from and to locations for each supplier were be returned in supplier specific queries i.e. two prices instead of one.

matt-crowson-moj:
Changes:

- This change resolves an issue whereby we were seeing multiple prices for a standard journey in the move view when there should only have been one.  The is the result of there being multiple prices at a given from and to location for both suppliers.
- Also added a bit more logging (unrelated to this particular fix.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
@@ -231,6 +231,8 @@ class HtmlController(@Autowired val moveService: MoveService, @Autowired val jou
 
   @GetMapping(SEARCH_JOURNEYS_URL)
   fun searchJourneys(model: ModelMap): Any {
+    logger.info("getting search journey")
+
     val effectiveYear = model.getEffectiveYear()
 
     model.addAttribute("form", SearchJourneyForm())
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
```diff
@@ -231,6 +231,8 @@ class HtmlController(@Autowired val moveService: MoveService, @Autowired val jou
 
   @GetMapping(SEARCH_JOURNEYS_URL)
   fun searchJourneys(model: ModelMap): Any {
+    logger.info("getting search journey")
```
matt-crowson-moj, line 234: Added a bit more (missing) logging to the controllers.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/move/MoveQueryRepositoryTest.kt
```diff
@@ -76,58 +78,91 @@ internal class MoveQueryRepositoryTest {
       )
     )
 
-    moveRepository.save(standardMove)
+    priceRepository.save(
```
matt-crowson-moj, line 81: This is duplicates the price but for GEOAmey to recreated and test the issue is resolved.

Thource (approved):

PR #232, opened 2021-03-11
P4-2696 changes to (hopefully) fix a session related issue on the choose supplier screen

matt-crowson-moj:
Changes:

So I have spent a bit of time trying to get to the bottom of this issue.  From what I can gather it only happens on the choose supplier screen.  I am not able to recreate it in my local development environment.

Looking at the code there may (possibly) be a race condition, this is an educated guess at this point.  So I have moved some of the logic closer to where it is actually needed/used.  The logic in question adds the chosen supplier to the users session (which ultimately gets persisted to the database session) and is then used straight after that by doing a redirect having chosen the supplier.  My suspicion is the time to persist it to the DB session and the point it is next used (it pulls it back from the session) after the redirect is the issue.

So as mentioned it is an educated guess change at this point in time given I cannot recreate it.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
@@ -18,6 +18,7 @@ import org.springframework.web.servlet.ModelAndView
 import org.springframework.web.servlet.mvc.support.RedirectAttributes
 import org.springframework.web.servlet.view.RedirectView
 import org.springframework.web.util.UriComponentsBuilder
+import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
 import uk.gov.justice.digital.hmpps.pecs.jpc.constraint.ValidJourneySearch
 import uk.gov.justice.digital.hmpps.pecs.jpc.constraint.ValidMonthYear
 import uk.gov.justice.digital.hmpps.pecs.jpc.controller.HtmlController.Companion.DATE_ATTRIBUTE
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
```diff
@@ -68,6 +73,8 @@ class HtmlController(@Autowired val moveService: MoveService, @Autowired val jou
     logger.info("chosen supplier Serco")
 
     model.addAttribute(SUPPLIER_ATTRIBUTE, Supplier.SERCO)
+    model.addAttribute(DATE_ATTRIBUTE, timeSource.startOfMonth())
```
matt-crowson-moj, line 76: One of the key change here...

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlController.kt
```diff
@@ -76,9 +83,13 @@ class HtmlController(@Autowired val moveService: MoveService, @Autowired val jou
     logger.info("chosen supplier GEOAmey")
 
     model.addAttribute(SUPPLIER_ATTRIBUTE, Supplier.GEOAMEY)
+    model.addAttribute(DATE_ATTRIBUTE, timeSource.startOfMonth())
```
matt-crowson-moj, line 86: One of the key change here.  This was being added in the filter....

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/filter/ChooseSupplierFilter.kt
```diff
@@ -22,16 +22,13 @@ class ChooseSupplierFilter(private val timeSource: TimeSource) : Filter {
...
-    val session = req.session
+    val session = (request as HttpServletRequest).session
 
-    when (session.getAttribute("supplier")) {
+    when (session.getAttribute(HtmlController.SUPPLIER_ATTRIBUTE)) {
       null -> {
-        logger.info("no supplier present in the session, redirecting to choose supplier")
+        logger.info("no supplier present in the session, redirecting to '${HtmlController.CHOOSE_SUPPLIER_URL}'")
 
-        session.setAttribute("date", timeSource.date().withDayOfMonth(1))
```
matt-crowson-moj, line 33: One of the key changes here.  Removed this from the filter and put logic into the Controller.

Thource (approved):

PR #234, opened 2021-03-16
P4-2710 initial work for putting in WireMock so we can fake the BaSM API calls when running locally.

matt-crowson-moj:
Introducing stubbed BaSM API for local development.  This is just to support calls for locations based on their NOMIS agency id.

### DIFF
--- a/wiremock-docker/__files/basm-api/locations/COURT2.json
+++ b/wiremock-docker/__files/basm-api/locations/COURT2.json
@@ -1,13 +1,13 @@
 {
   "data": [
     {
-      "id": "7bad3789-0d49-44dc-bb8f-4ff32414fa0b",
+      "id": "2bad3789-0d49-44dc-bb8f-4ff32414fa0b",
       "type": "locations",
       "attributes": {
         "key": "court2",
         "title": "Court Two",
         "location_type": "court",
-        "nomis_agency_id": "COURT1",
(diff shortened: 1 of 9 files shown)


wiremock-docker/__files/basm-api/locations/COURT2.json
```diff
@@ -0,0 +1,44 @@
+{
+  "data": [
+    {
+      "id": "7bad3789-0d49-44dc-bb8f-4ff32414fa0b",
+      "type": "locations",
+      "attributes": {
+        "key": "court2",
+        "title": "Court Two",
+        "location_type": "court",
+        "nomis_agency_id": "COURT1",
```
Thource, line 10: 2?

wiremock-docker/__files/basm-api/locations/COURT2.json
```diff
@@ -0,0 +1,44 @@
+{
+  "data": [
+    {
+      "id": "7bad3789-0d49-44dc-bb8f-4ff32414fa0b",
```
Thource, line 4: Will non-unique IDs cause issues?

matt-crowson-moj: Nope as we won't use them, we only care about the location name.  I'll change them though anyway.

Thource (approved):

PR #238, opened 2021-03-22
P4-2701 changes to UI layer to display the NOMIS location name when mapping an agency without a Schedule 34 mapping.

matt-crowson-moj:
Changes:

Displays the NOMIS location name (if one can be found for the selected agency ID) on the add/update mappings screen to help the user when mapping a Schedule 34 location name.


src/main/resources/spreadsheets/nomis-locations.csv
```diff
@@ -1,4 +1,7 @@
 AGY_LOC_ID,DESCRIPTION,LONG_DESCRIPTION,AGENCY_LOCATION_TYPE,ACTIVE_FLAG
+COURT1,NOMIS Court One,NOMIS Court One,CRT,Y
```
matt-crowson-moj, line 2: These are faked locations for testing purposes.  They should be safe to leave in as nothing in NOMIS will ever map to them.  They can be removed one dev has been tested.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/MapFriendlyLocationController.kt
```diff
@@ -68,7 +90,7 @@ class MapFriendlyLocationController(private val service: MapFriendlyLocationServ
 
     service.mapFriendlyLocation(form.agencyId, form.locationName, LocationType.valueOf(form.locationType))
 
-    redirectAttributes.addFlashAttribute("flashAttrMappedLocationName", form.locationName)
+    redirectAttributes.addFlashAttribute("flashAttrMappedLocationName", form.locationName.toUpperCase())
```
matt-crowson-moj, line 93: Small tweak as this are upper-cased when saved.  I guess ideally should come from the service but didn't seem worth it.

Thource (approved):

PR #241, opened 2021-03-25
P4-2721 add flyway baseline to easily migrations going forwards.

matt-crowson-moj:
Changes:

The purpose of this change is to cleanup the way we handle DB changes.  Flyway support was never integrated with a baselined version of the DB objects and as such means adding new migrations difficult and fragile.

There is some manual intervention needed on the DB's to support this change. Before releasing to any environment the Flyway history table needs cleaning and a baseline insert manually applied that matches with the file in this commit.


src/main/resources/application.yml
```diff
@@ -46,7 +46,7 @@ spring:
           issuer-uri: ${HMPPS_AUTH_BASE_URI}/issuer
 
   flyway:
-    locations: classpath:/db/migration/ddl,classpath:/db/migration/data/all
```
matt-crowson-moj, line 49: We don't need this at the moment as this change includes a baseline SQL file.

src/main/resources/db/migration/ddl/V1__baseline.sql
```diff
@@ -0,0 +1,128 @@
...
+
+create table profiles
+(
+    profile_id varchar(255) not null,
+    person_id  varchar(255) not null,
+    updated_at timestamp    not null,
+    constraint PROFILES_PK primary key (profile_id)
+);
+
+create table shedlock
```
matt-crowson-moj, line 117: So I have included this in here as well.  I think this OK for now.

Thource (approved):

PR #246, opened 2021-03-30
P4-2728 moving report imports to run later in the day.

matt-crowson-moj:
**Changes:**

Changing morning report imports to run at 5:30am (instead of 4am) to leave sufficient gap for when the files are available for actual import.

**Why was this needed?**

There was a gap of an hour between the files being uploaded (by a separate process) and then being downloaded by the application. When the clocks changed both processes ran at the same time.  This was due to one process running in UTC and the other in BST.

### DIFF
--- a/helm_deploy/values-preprod.yaml
+++ b/helm_deploy/values-preprod.yaml
@@ -26,7 +26,7 @@ env:
   JAVA_OPTS: "-Xmx2048m"
   HMPPS_AUTH_BASE_URI: "https://sign-in-preprod.hmpps.service.justice.gov.uk/auth"
   HMPPS_AUTH_REDIRECT_BASE_URI: "https://calculate-journey-variable-payments-preprod.apps.live-1.cloud-platform.service.justice.gov.uk"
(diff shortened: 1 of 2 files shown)


helm_deploy/values-preprod.yaml
```diff
@@ -26,7 +26,7 @@ env:
   JAVA_OPTS: "-Xmx2048m"
   HMPPS_AUTH_BASE_URI: "https://sign-in-preprod.hmpps.service.justice.gov.uk/auth"
   HMPPS_AUTH_REDIRECT_BASE_URI: "https://calculate-journey-variable-payments-preprod.apps.live-1.cloud-platform.service.justice.gov.uk"
-  CRON_IMPORT_REPORTS: "0 0 4 * * ?"
+  CRON_IMPORT_REPORTS: "0 30 5 * * ?"
```
matt-crowson-moj, line 29: Note these times are using Spring CRON (not standard CRON).  Spring being Spring it does it slightly differently and allows for seconds hence the extra 0 at the front.

Thource (approved):

PR #263, opened 2021-05-11
P4-2833 providing the ability to automatically backdate locations from BaSM via the CLI tooling.

matt-crowson-moj:
**Changes:**

- Adding command to the CLI so we can automatically backdate a location from BaSM.
- Some small tweaks to the WireMock files just to make local manual testing a bit easier.

### DIFF
--- a/wiremock-docker/mappings/basm_api_location_MULTIPLE_ON_DATE.json
+++ b/wiremock-docker/mappings/basm_api_location_MULTIPLE_ON_DATE.json
@@ -1,6 +1,11 @@
 {
   "request" : {
-    "url" : "/api/reference/locations?filter%5Bcreated_at%5D=2021-05-05&per_page=100",
+    "urlPathPattern": "/api/reference/locations",
+    "queryParameters" : {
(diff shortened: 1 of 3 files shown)


wiremock-docker/mappings/basm_api_location_MULTIPLE_ON_DATE.json
```diff
@@ -1,6 +1,11 @@
 {
   "request" : {
-    "url" : "/api/reference/locations?filter%5Bcreated_at%5D=2021-05-05&per_page=100",
+    "urlPathPattern": "/api/reference/locations",
+    "queryParameters" : {
+      "filter%5Bcreated_at%5D" : {
+        "matches" : "^\\d{4}-\\d{2}-\\d{2}$"
```
matt-crowson-moj, line 6: This is a crude regex but fine for local testing.  The point at which it is actually called the application will only ever provide a valid date.  It doesn't really care what the date is.  The response will always default to the current date.  So WireMock can pull through query param values but sadly not in the format used by BaSM.

PR #264, opened 2021-05-14
P4-2833 removing redudant date parsing from location call on return fom call to BaSM endpoint.

matt-crowson-moj:
**Changes:**

This change removes the unnecessary parsing of the **created_at** date/timestamp in the response from the call to BaSM locations reference data.  We don't actually care or use it.

### DIFF
--- a/wiremock-docker/__files/basm-api/locations/MULTIPLE.json
+++ b/wiremock-docker/__files/basm-api/locations/MULTIPLE.json
@@ -18,7 +18,7 @@
         "latitude": null,
         "longitude": null,
         "disabled_at": null,
-        "created_at": "{{now format='yyyy-MM-dd'}}"
+        "created_at": "{{now}}"
       },
       "relationships": {
         "suppliers": {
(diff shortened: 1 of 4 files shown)


wiremock-docker/__files/basm-api/locations/MULTIPLE.json
```diff
@@ -70,7 +70,7 @@
         "latitude": null,
         "longitude": null,
         "disabled_at": null,
-        "created_at": "{{now format='yyyy-MM-dd'}}"
+        "created_at": "{{now}}"
```
Thource, line 73: Do these want removing?

Thource (approved):

wiremock-docker/__files/basm-api/locations/MULTIPLE.json
matt-crowson-moj: Left in attempt to be consistent with the response.  Though saying that the same should apply to all the other JSON files really...

PR #270, opened 2021-05-19
P4-2833 resolve issue with potential location conflicts from poor (NOMIS) data.

matt-crowson-moj:
**Changes:**

When attempting mapping on production via the CLI an unexpected problem showed up preventing the mapping of new locations.  One location from BaSM matched two locations in CJVP.  This is actually poor location data on the whole which cannot be resolved at the code level.  As a result all we can do is alert to the fact so it can be raised with the end users.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/AutomaticLocationMappingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/AutomaticLocationMappingService.kt
@@ -15,7 +15,8 @@ class AutomaticLocationMappingService(
   private val basmClientApi: BasmClientApiService,
   private val locationRepository: LocationRepository,
   private val timeSource: TimeSource,
-  private val auditService: AuditService
+  private val auditService: AuditService,
+  private val monitoringService: MonitoringService
 ) {
 
   private val logger = LoggerFactory.getLogger(javaClass)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/AutomaticLocationMappingService.kt
```diff
@@ -15,19 +15,27 @@ class AutomaticLocationMappingService(
...
+              auditService.create(AuditableEvent.autoMapLocation(newLocation))
 
-            logger.info("Automatically mapped new location: agency ID '${it.agencyId}', name '${it.name}' and type '${it.locationType.label}'")
+              logger.info("Automatically mapped new location: agency ID '${it.agencyId}', name '${it.name}' and type '${it.locationType.label}'")
+            }
+          }
+          1 -> logger.info("Location with agency id '${it.agencyId}' and name '${it.name}' already exists, no mapping required.")
+          else -> {
+            logger.warn("Multiple locations for agency id '${it.agencyId}' and location name '${it.name}', raise with end users.")
+            monitoringService.capture("Multiple locations for agency id '${it.agencyId}' and location name '${it.name}', raise with end users.")
```
matt-crowson-moj, line 38: So raising an alert so we can at least make the users aware so they can look into it.  This is the only way we will know as it is an automated process.

Thource (approved):

PR #272, opened 2021-05-21
P4-2895 migration to backfill location history audit events.  The audting functionality came after some of the locations were there.

matt-crowson-moj:
**Changes:**

A new migration has been added to backfill location history where there wasn't any previously.  This was due to the fact they were added to the system prior to the auditing functionality being added. The reason for doing this is so we can display something on the new location history change, otherwise it would have been blank and confusing to the user.

The new migration is using a Postgres specific function for UUID generation.  This in turn has the knock on effect of not working with an in memory DB. So to get around that I have introduced a Postgres TestContainer so we can run them fully.  This has meant splitting out into an integration test which at time of writing does not run as part of the CI pipeline.  This is on my list of things to do.


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/MigrationsTest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+@SpringBootTest
+@ActiveProfiles("test")
+@ContextConfiguration(classes = [TestConfig::class])
+@TestPropertySource(properties = ["spring.flyway.enabled=true", "spring.flyway.baseline-on-migrate=true", ])
+internal class MigrationsTest {
+
+  @Test
+  internal fun `migrations run without any errors when running against a Postgres DB instance`() {
+    // no assertion necessary test will fail if any of the migrations fail to run e.g. malformed SQL
```
matt-crowson-moj, line 21: We are now making sure the migrations run without any errors.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/util/PostgresContainer.kt
```diff
@@ -0,0 +1,35 @@
...
+import java.io.IOException
+import java.net.ServerSocket
+
+object PostgresContainer {
+
+  val instance: PostgreSQLContainer<Nothing>? by lazy { startPostgresqlContainer() }
+
+  fun startPostgresqlContainer(): PostgreSQLContainer<Nothing>? =
+    if (checkPostgresRunning().not()) {
+      PostgreSQLContainer<Nothing>("postgres:10.16").apply {
```
matt-crowson-moj, line 14: This is using the same version of PG as prod (and the docker compose file).

Thource (approved):

PR #276, opened 2021-05-26
P4-2896 changes to improve the performance when reading the audit history events.

matt-crowson-moj:
**Changes**:

The is the first attempt to improve the read time of the audit history events.  I don't want to do too much at this point so only making minor changes.  Adds an new index and removes some unnecessary string santisations.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/SupplierPricingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/SupplierPricingService.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.pecs.jpc.service
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.pecs.jpc.auditing.AuditEvent
 import uk.gov.justice.digital.hmpps.pecs.jpc.auditing.AuditEventType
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/SupplierPricingService.kt
```diff
@@ -106,11 +107,15 @@ class SupplierPricingService(
...
-  fun priceHistoryForJourney(supplier: Supplier, fromAgencyId: String, toAgencyId: String) =
-    auditService.auditEventsByType(AuditEventType.JOURNEY_PRICE)
+  fun priceHistoryForJourney(supplier: Supplier, fromAgencyId: String, toAgencyId: String): Set<AuditEvent> {
+    val sanitisedFrom = sanitised(fromAgencyId)
+    val sanitisedTo = sanitised(toAgencyId)
+
+    return auditService.auditEventsByType(AuditEventType.JOURNEY_PRICE)
       .associateWith { PriceMetadata.map(it) }
-      .filterValues { it.supplier == supplier && sanitised(it.fromNomisId) == sanitised(fromAgencyId) && sanitised(it.toNomisId) == sanitised(toAgencyId) }
+      .filterValues { it.supplier == supplier && it.fromNomisId == sanitisedFrom && it.toNomisId == sanitisedTo }
```
matt-crowson-moj, line 116: So this would happen loads of times given there are 25000 + price audit events.  We don't need to sanitise each row.

Thource (approved):

PR #284, opened 2021-06-15
chore: Use HMPPS Orb owasp check and post to pecs dev channel

Nimphal:
Updates the OWASP job to use the one from the HMPPS orb and specifies it should post to the pecs_dev slack channel on failure.


matt-crowson-moj (approved):
LGTM

PR #291, opened 2021-06-25
P4-2906 first cut of adding integration testing.  This is a WIP.

matt-crowson-moj:
**Changes:**

This is the first cut of adding some level of integration testing to the application.  It is pretty basic at the moment but will hopefully provide a baseline to be improved upon.

At the moment this relies on the user to run the integration tests (see the readme file for details).  Ideally we would want this to go into a CI pipeline somewhere.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,33 +3,60 @@
 version: '3.1'
 
 services:
+  calculate-journey-variable-payments:
+    restart: on-failure:3
+    image: quay.io/hmpps/calculate-journey-variable-payments:latest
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -3,33 +3,60 @@
 version: '3.1'
 
 services:
+  calculate-journey-variable-payments:
+    restart: on-failure:3
```
matt-crowson-moj, line 7: This is a workaround that needs addressing.  The OAuth2 service takes too long to start so this service fails to start initially.

Thource (approved):

PR #294, opened 2021-07-01
P4 2906 add integration tests to ci pipeline

matt-crowson-moj:
**Changes:**

This change introduces the integration tests as part of the CI pipeline.

Assuming the unit tests pass this part of the workflow will then be executed.  If this passes then we are done and it can be deployed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,26 +50,18 @@ jobs:
       - attach_workspace:
           at: ~/app
       - run:
-          name: Install missing OS dependency
-          command: sudo apt-get install wget
-#      - run:
-#          name: Install dockerize
-#          command: wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
-#          environment:
-#            DOCKERIZE_VERSION: v0.6.1
-      - run:
           name: Wait for auth
           command: dockerize -wait http://localhost:9090/auth/health -timeout 1m
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -5,15 +5,17 @@ orbs:
 
 jobs:
   validate:
-    executor: hmpps/java
+    executor:
+      name: hmpps/java
+      tag: "11.0"
```
matt-crowson-moj, line 10: Whilst not strictly necessary I have set this to 11.0 in case the ORB default happens to change.  This also keeps it in line with the base image used below **cimg/openjdk:11.0**.

.circleci/config.yml
```diff
@@ -22,6 +24,68 @@ jobs:
...
+      - run:
+          name: Get wiremock
+          command: curl -o wiremock.jar https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-standalone/2.27.2/wiremock-standalone-2.27.2.jar
+      - run:
+          name: Run wiremock
+          command: java -jar wiremock.jar --port 9999
+          background: true
+      - run:
+          name: Run the java app.
+          command: env spring.profiles.active=dev java -Xmx1024m -jar ~/app/build/libs/calculate-journey-variable-payments-*.jar
```
matt-crowson-moj, line 64: Not 100% sure why I need to specify `~/app/` here for jar execution as the folder is added at the top of the steps: above.

src/main/resources/application-dev.yml
```diff
@@ -7,3 +7,22 @@ spring:
   flyway:
     enabled: true
     locations: classpath:/db/migration
+
+APP_DB_URL: "jdbc:postgresql://127.0.0.1:5433/jpc?reWriteBatchedInserts=true&user=jpc&password=letmein"
```
matt-crowson-moj, line 11: Brought these in here so the profile can be specified when running the integration tests.

This profile also adds some test data via Flywaydb migrations.

Nimphal (approved):

PR #295, opened 2021-07-02
P4-2962 initial commit for integration tests for add and update price.

matt-crowson-moj:
**Changes:**

First cut of adding integration test coverage for the add and update price scenarios.

_Note:  It isn't quite right as there is currently an issue with some drop down JavaScript not firing for location name look-ups so currently I am going direct to a URL instead of via the screen it should._

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManagePricesTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManagePricesTest.kt
@@ -1,0 +1,92 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.integration
+
+import org.fluentlenium.core.annotation.Page
+import org.junit.jupiter.api.MethodOrderer.OrderAnnotation
+import org.junit.jupiter.api.Order
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestMethodOrder
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.AddPricePage
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.ChooseSupplierPage
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.DashboardPage
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.JourneysForReviewPage
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.ManageJourneyPricePage
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SelectMonthYearPage
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.UpdatePricePage
+import uk.gov.justice.digital.hmpps.pecs.jpc.price.Money
+import uk.gov.justice.digital.hmpps.pecs.jpc.price.Supplier
+import java.time.LocalDate
+import java.time.Month
+import java.time.Year
+
+@TestMethodOrder(OrderAnnotation::class)
+internal class ManagePricesTest : IntegrationTest() {
+
+  @Page
+  private lateinit var dashboardPage: DashboardPage
+
+  @Page
+  private lateinit var chooseSupplierPage: ChooseSupplierPage
+
+  @Page
+  private lateinit var selectMonthYearPage: SelectMonthYearPage
+
+  @Page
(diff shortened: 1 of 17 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManagePricesTest.kt
```diff
@@ -0,0 +1,92 @@
...
+  private lateinit var addPricePage: AddPricePage
+
+  @Page
+  private lateinit var manageJourneyPricePage: ManageJourneyPricePage
+
+  @Page
+  private lateinit var updatePricePage: UpdatePricePage
+
+  @Test
+  @Order(1)
```
matt-crowson-moj, line 46: Due to the fact these are integration tests (i.e. data is persisted across tests) the tests are being ordered so we can add then add then update the price for the same journey.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/pages/ManageJourneyPricePage.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages
+
+import org.fluentlenium.core.annotation.PageUrl
+
+@PageUrl("http://localhost:8080/search-journeys")
+class ManageJourneyPricePage : ApplicationPage() {
+
+  // Note: this is bypassing the actual search journey page due to not liking the JavaScript in the page where the fields are added by the 3rd party JS lib.
```
matt-crowson-moj, line 8: So at the moment this is a workaround.  Need to look into this some more.  We have something for starters so I am happy with that.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManagePricesTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    chooseSupplierPage.isAt()
+    chooseSupplierPage.choose(Supplier.GEOAMEY)
+
+    dashboardPage.isAt()
+    dashboardPage.navigateToManageJourneyPrice()
+
+    manageJourneyPricePage.isAt()
+    manageJourneyPricePage.findJourneyToManagePrice("PRISON1", "PRISON2")
+
+    updatePricePage.isAtPricePageForJourney("PRISON1", "PRISON2")
```
matt-crowson-moj, line 89: Could possibly do with an assertion to check the price before updating i.e. it should be £10.00.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManagePricesTest.kt
Nimphal: Making tests run in a specific order is a bit of a smell, ideally they'd be completely independent. I would say it's perfectly fine to duplicate the add price bit in the update price test - it documents that a price has to have been added before it can be updated, so it acts as a documentation of how the system works.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/pages/AddPricePage.kt
```diff
@@ -0,0 +1,26 @@
...
+  private lateinit var submit: FluentWebElement
+
+  fun isAtPricePageForJourney(fromAgencyId: String, toAgencyId: String) {
+    this.isAt("$fromAgencyId-$toAgencyId")
+  }
+
+  fun addPriceForJourney(fromAgencyId: String, toAgencyId: String, amount: Money) {
+    this.isAtPricePageForJourney(fromAgencyId, toAgencyId)
+    price.fill().withText(amount.toString())
+    submit.click()
```
Nimphal, line 24: What does success look like here? In the tests I did not see us check for a confirmation page or something - what would happen if this errorred out, would the tests still pass?

Nimphal (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/pages/AddPricePage.kt
matt-crowson-moj: Yes we can...same goes for update.

image

PR #296, opened 2021-07-05
P4-2962 refactoring integration tests around price management.

matt-crowson-moj:
**Changes:**

- Additional assertion added when 'adding' a price to confirm the price has actually been added.
- More work is needed around the update step due to an issue with JavaScript execution not working with HtmlUnit via Selenium.
- Refactoring the integrations tests to be more Fluent based.
- Migration tests no longer needed as we get migration coverage as a result of running the integration tests.


docker-compose-cjvp.yml
```diff
@@ -0,0 +1,20 @@
+version: '3.1'
```
matt-crowson-moj, line 1: Pulled this out of the main compose file so we can run as and when needed locally e.g. running the whole suit from one command line operation.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManagePricesTest.kt
```diff
@@ -44,49 +45,69 @@ internal class ManagePricesTest : IntegrationTest() {
...
-    addPricePage.addPriceForJourney("PRISON1", "PRISON2", Money(1000))
+    journeysForReviewPage
+      .isAtPage<JourneysForReviewPage>()
+      .addPriceForJourney("PRISON1", "PRISON2")
+
+    addPricePage
+      .isAtPricePageForJourney("PRISON1", "PRISON2")
+      .addPriceForJourney("PRISON1", "PRISON2", Money(1000))
+
+    journeysForReviewPage
```
matt-crowson-moj, line 81: New assertion added here to confirm prices has indeed been added.

Nimphal (approved):

PR #297, opened 2021-07-09
P4-2962 using chromedriver to resolve issue with JavaScript not working. Price add and update pages now following correct path.

matt-crowson-moj:
**Changes:**

The changes in this PR now mean we are using chromedriver to drive the integration tests.  This resolves the issue around JavaScript not working (with HtmlUnit).

Both the add and update price screens now follow the correct flows through the application.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -56,6 +56,9 @@ dependencies {
   testImplementation("com.squareup.okhttp3:okhttp:4.9.1")
   testImplementation("org.testcontainers:postgresql:1.15.3")
 
+  testImplementation("org.seleniumhq.selenium:selenium-support:3.141.59")
+  testImplementation("org.seleniumhq.selenium:selenium-chrome-driver:3.141.59")
+
   runtimeOnly("org.flywaydb:flyway-core:7.10.0")
   runtimeOnly("com.h2database:h2")
   runtimeOnly("org.postgresql:postgresql:42.2.22")
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   hmpps: ministryofjustice/hmpps@3.2
+  browser-tools: circleci/browser-tools@1.1.3
 
 jobs:
   validate:
@@ -32,7 +33,7 @@ jobs:
 
   integration_tests:
     docker:
-      - image: cimg/openjdk:11.0
+      - image: cimg/openjdk:11.0-browsers
         environment:
           _JAVA_OPTIONS: -Xmx1024m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -XX:ParallelGCThreads=2 -Djava.util.concurrent.ForkJoinPool.common.parallelism=2 -Dorg.gradle.daemon=false
       - image: quay.io/hmpps/hmpps-auth:latest
@@ -49,6 +50,7 @@ jobs:
       - checkout
       - attach_workspace:
           at: ~/app
+      - browser-tools/install-browser-tools
       - run:
           name: Wait for auth
           command: dockerize -wait http://localhost:9090/auth/health -timeout 1m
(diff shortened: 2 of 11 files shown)


build.gradle.kts
```diff
@@ -56,6 +56,10 @@ dependencies {
   testImplementation("com.squareup.okhttp3:okhttp:4.9.1")
   testImplementation("org.testcontainers:postgresql:1.15.3")
 
+  testImplementation("org.seleniumhq.selenium:selenium-support:3.141.59")
+  testImplementation("org.seleniumhq.selenium:selenium-chrome-driver:3.141.59")
+  testImplementation("org.seleniumhq.selenium:selenium-firefox-driver:3.141.59")
```
matt-crowson-moj, line 61: I'll remove this (Firefox driver) in a separate PR.

thomasleese (commented):
The change makes sense to me, but I’ll leave it to someone else to approve!

I was just wandering if it was worth splitting the commit up into smaller ones? I’m thinking perhaps the introduction of chromedriver in one and then the utilisation of it in another?

matt-crowson-moj:
> The change makes sense to me, but I’ll leave it to someone else to approve!
> 
> I was just wandering if it was worth splitting the commit up into smaller ones? I’m thinking perhaps the introduction of chromedriver in one and then the utilisation of it in another?

Thanks Tom but in short no.  The changes are dependent on the introduction of the Chromedriver.  The CI pipeline needs the change at the same time or the build will fail and can't be merged.

thomasleese:
> The changes are dependent on the introduction of the Chromedriver. The CI pipeline needs the change at the same time or the build will fail and can't be merged.

I was imagining that it's okay if the build _would_ fail on an individual commit in the PR, as long as the final commit and merge commit into main build correctly. I tend to prefer lots of smaller commits to make it easier to do `git blame` or `git bisect` in the future or something like that. Either way, I don't really mind too much.

I have thought of a second question though, probably not for this PR, but to be noted down somewhere, is it worth us switching all the tests over in this repo to run using `chromedriver`? It might avoid some potential confusion where it's unclear how the tests are being run, and keeps things nice and consistent. I imagine a downside of this is that the tests take longer to run.

Further down the line, it might also allow us to do something clever in the future if we know the tests run under `chromedriver` we could try running them under `geckodriver` as well to be aware of any potential browser differences?

matt-crowson-moj:
> > The changes are dependent on the introduction of the Chromedriver. The CI pipeline needs the change at the same time or the build will fail and can't be merged.
> 
> I was imagining that it's okay if the build _would_ fail on an individual commit in the PR, as long as the final commit and merge commit into main build correctly. I tend to prefer lots of smaller commits to make it easier to do `git blame` or `git bisect` in the future or something like that. Either way, I don't really mind too much.
> 
> I have thought of a second question though, probably not for this PR, but to be noted down somewhere, is it worth us switching all the tests over in this repo to run using `chromedriver`? It might avoid some potential confusion where it's unclear how the tests are being run, and keeps things nice and consistent. I imagine a downside of this is that the tests take longer to run.
> 
> Further down the line, it might also allow us to do something clever in the future if we know the tests run under `chromedriver` we could try running them under `geckodriver` as well to be aware of any potential browser differences?

All tests bar one are now using Chromedriver.  There is one test using a different driver (for now) due to testing a file download.  This is an acceptable trade off for now.  The thing to be aware of here this has only recently been added (integration testing that is).  So with regards to different browsers sure, certainly that is something to consider going forwards. The main driver here however was getting a framework in place to enable us to do that in future.

Thource (approved):

thomasleese:
> All tests bar one are now using Chromedriver. There is one test using a different driver (for now) due to testing a file download. This is an acceptable trade off for now. The thing to be aware of here this has only recently been added (integration testing that is). So with regards to different browsers sure, certainly that is something to consider going forwards. The main driver here however was getting a framework in place to enable us to do that in future.

Oh sorry, I'd misunderstood and thought this PR was about adding Chromedriver for one of the tests, but I see it's actually about adding Chromedriver for all the tests except the one which doesn't work with it!

PR #299, opened 2021-07-12
P4-2694 new integration test coverage for add and update NOMIS locations.

matt-crowson-moj:
**Changes:**

This PR adds integration test coverage for the add and update NOMIS locations.

_note: some refactorings are included as part of this PR to reduce the amount of boiler plate in the actual tests which will hopefully make them a little less verbose but still clear enough to follow and understand the intentions._

### DIFF
--- a/src/main/resources/templates/add-location.html
+++ b/src/main/resources/templates/add-location.html
@@ -74,7 +74,7 @@
             </strong>
         </div>
         <div class="flex items-center">
-            <button class="govuk-button mv3 mr3" data-module="govuk-button">
+            <button id="confirm-save-location" class="govuk-button mv3 mr3" data-module="govuk-button">
                 Confirm and save
             </button>
             <a href="/journeys" class="govuk-link">Cancel</a>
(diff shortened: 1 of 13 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/pages/Pages.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages
+
+sealed class Pages<T : ApplicationPage>(val page: Class<T>) {
```
matt-crowson-moj, line 3: I would have liked to use an enum for this but you can not use type erasure like this in Kotlin enums hence the Sealed class.

src/main/resources/templates/add-location.html
```diff
@@ -74,7 +74,7 @@ <h1 class="govuk-heading-xl mv2" th:inline="text">Map location</h1>
             </strong>
         </div>
         <div class="flex items-center">
-            <button class="govuk-button mv3 mr3" data-module="govuk-button">
+            <button id="confirm-save-location" class="govuk-button mv3 mr3" data-module="govuk-button">
```
matt-crowson-moj, line 77: Adding ID's for so elements can easily be found by Selenium.

Thource (approved):

thomasleese (approved):

PR #304, opened 2021-07-15
P4-2967 this is a work in progress - it is a code based migration to assist with development and integration tests.

matt-crowson-moj:
**Changes:**

A code based migration to assist with both development and integration tests. At the moment it will generate standard moves only for the day on which it is run.  I will be adding test data for all move types.

This enables me to easily control the dates of the individual moves so I can determine what things should like for integration tests purposes given they are very much time based.

As a bit a plus this also means we can generate current data (i.e. for the current month) when developing locally.

Attached screenshot is an example of what you would currently see running the application locally:

image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -35,6 +35,7 @@ dependencies {
     }
   }
 
+  implementation("org.flywaydb:flyway-core:7.10.0")
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
   implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
@@ -59,7 +60,6 @@ dependencies {
   testImplementation("org.seleniumhq.selenium:selenium-support:3.141.59")
   testImplementation("org.seleniumhq.selenium:selenium-chrome-driver:3.141.59")
 
-  runtimeOnly("org.flywaydb:flyway-core:7.10.0")
   runtimeOnly("com.h2database:h2")
   runtimeOnly("org.postgresql:postgresql:42.2.22")
 }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/db/migration/data/dev/R__2_5_Integration_test_data.kt
```diff
@@ -0,0 +1,265 @@
...
+      event.occurredAt,
+      event.recordedAt,
+      event.supplier?.name,
+      event.type,
+      event.updatedAt
+    )
+  }
+}
+
+private val today = LocalDate.now()
```
matt-crowson-moj, line 112: This is one of the key elements here.  This enables fine grained control of the dates and times of the moves, journeys and events.

Thource (approved):

build.gradle.kts
```diff
@@ -35,6 +35,7 @@ dependencies {
     }
   }
 
+  implementation("org.flywaydb:flyway-core:7.10.0")
```
thomasleese, line 38: Out of interest what does this change from `runtimeOnly` to `implementation` do? (Is it worth having a separate commit for it to explain the reasoning?)

matt-crowson-moj: No a separate commit is not necessary for this.  I cannot write the code migration without it.  It was only available at runtime prior to this

This is standard stuff and probably more down to experience with Java/Kotlin/Gradle.

thomasleese: I imagined it might be nice to have the commit split up into something like this:

```
Make flyway-core an implementation dependency

This is required to write database migrations in code. Previously it was
only available at runtime.
```

```
Add migration to assist with development

...[PR description here]...
```

Personally, I tend to prefer having lots of smaller commits as is it makes it easier to look back at the changes without the context of it being worked on right now. I think it's especially helpful for other developers who want to understand what's changed in the repo. I find it makes using tools like [`git bisect`](https://git-scm.com/docs/git-bisect) more successful as well.

I quite like [this article on commit messages](https://chris.beams.io/posts/git-commit/) which explains it much better than I could!

PR #309, opened 2021-07-26
P4-3042 fixing bug whereby multiple journey prices (for same journey different year) being returned when trying to update. Was missing use of a query parameter.

matt-crowson-moj:
**Changes:**

Small change where a query parameter was not being used to limit the journey price returned during update price, getting multiple instead of a single price.  This has only just been noticed in preprod testing (not in production) due to price uplifts being run in preprod.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/SupplierPricingService.kt
```diff
@@ -50,7 +50,7 @@ class SupplierPricingService(
     effectiveYear: Int
```
matt-crowson-moj, line 50: The parameter was being passed in but wasn't actually being used.

Thource (approved):

PR #313, opened 2021-07-28
P4 3011 refactoring price adjustments for auditing

matt-crowson-moj:
**Changes:**

This PR is mainly refactoring in preparation for adding additional auditing at the individual price level.

Some of the renaming of classes/objects/entities has come about after a conversation with the Commercial team which has given us better understanding of their needs with respect to annual (bulk) price changes.  For example 'price uplifting' is just one type of price change/adjustment but it is a price adjustment.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/price/AnnualPriceAdjuster.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/price/AnnualPriceAdjuster.kt
@@ -5,17 +5,17 @@ import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
 import kotlin.streams.asSequence
 
 /**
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/price/AnnualPriceAdjuster.kt
```diff
@@ -0,0 +1,96 @@
...
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
+import kotlin.streams.asSequence
+
+/**
+ * Domain level service to perform the annual price adjustments for a supplier.
+ */
+@Component
+class AnnualPriceAdjuster(
```
matt-crowson-moj, line 11: Used to be the PriceUplifter. Now renamed.

Thource (approved):

PR #318, opened 2021-07-30
P4 3011 tidy ups for bulk pricing

matt-crowson-moj:
**Changes:**

- Updating the README around recent pricing changes.
- Removing now redundant functionality e.g. old bulk price change code.
- Locking down uplifts so they cannot be performed prior to the current effective year.  This is not something we should allow so putting guards in place to prevent (accidents).


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/commands/BulkPriceUpdateCommands.kt
```diff
@@ -1,29 +1,14 @@
...
 
 @ShellComponent
 class BulkPriceUpdateCommands(
-  private val bulkPricesService: BulkPricesService,
   private val annualPriceAdjustmentsService: AnnualPriceAdjustmentsService
 ) {
-
-  private val logger = LoggerFactory.getLogger(javaClass)
-
-  @ShellMethod("Adds prices for the next effective year based on the current effective years prices and the supplied multiplier.")
```
matt-crowson-moj, line 18: Removed because this is no longer applicable.

Thource (approved):

PR #333, opened 2021-08-31
P4 3009 display price adjustment history feature

matt-crowson-moj:
**Changes:**

- Adding price adjustment history tab to the price adjust page/view.
- Bumping number of allowed decimal points from 3 to 4 for the price adjustment rate.

See example screenshot below:

image


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ViewAllMoveTypesTest.kt
```diff
@@ -25,6 +26,7 @@ import java.time.LocalDate
 import java.time.Month
 import java.time.Year
 
+@Disabled
```
matt-crowson-moj, line 29: @Thource FYI this is something to be aware of. I have disabled this whilst I continue to work on this feature.  As it stands I think we are OK.  I think the fault likes in the integration test data but this requires a bit of a rethink.  The issue is around moves/journeys that span months e.g. start on 31st and finish on the 1st of the next month.

Thource (approved):

PR #341, opened 2021-09-13
P4-3155 changes to include the application survey feeback link on all pages from the dashboard onwards.

matt-crowson-moj:
**Changes:**

Add an application wide feedback link to the bottom of all pages from the dashboard on-wards.  Note for development and pre-prod the links will not go anywhere.  This pulls in an env variable from the Kubernettes config (which you cannot see in this PR).

image

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/AnnualPriceAdjustmentsControllerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/AnnualPriceAdjustmentsControllerTest.kt
@@ -63,6 +63,8 @@ class AnnualPriceAdjustmentsControllerTest(@Autowired private val wac: WebApplic
       .andExpect { model { attribute("contractualYearStart", effectiveYearForDate(effectiveDate).toString()) } }
       .andExpect { model { attribute("contractualYearEnd", (effectiveYearForDate(effectiveDate) + 1).toString()) } }
       .andExpect { model { attributeExists("history") } }
+      // including check for the feedback URL
+      .andExpect { model { attribute("feedbackUrl", "#") } }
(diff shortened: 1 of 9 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/AnnualPriceAdjustmentsControllerTest.kt
```diff
@@ -63,6 +63,8 @@ class AnnualPriceAdjustmentsControllerTest(@Autowired private val wac: WebApplic
       .andExpect { model { attribute("contractualYearStart", effectiveYearForDate(effectiveDate).toString()) } }
       .andExpect { model { attribute("contractualYearEnd", (effectiveYearForDate(effectiveDate) + 1).toString()) } }
       .andExpect { model { attributeExists("history") } }
+      // including check for the feedback URL
```
matt-crowson-moj, line 66: There is no actual controller for the feedback URL it is effectively added to all existing controllers so including a simple check here.  In this case there is URL.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/HtmlControllerTest.kt
```diff
@@ -45,6 +50,16 @@ class HtmlControllerTest(@Autowired private val wac: WebApplicationContext) {
     mockSession.setAttribute("supplier", defaultSupplierSerco)
   }
 
+  @Test
+  internal fun `feedback URL is available upon navigation to the dashboard`() {
```
matt-crowson-moj, line 54: Including a simple check here where there is a feedback URL.

Thource (approved):

Thource (approved):

PR #343, opened 2021-09-14
P4 3147 refactoring integration test data

matt-crowson-moj:
**Changes:**

The fixture data for the integration tests was due a bit of a rethink having noticed some temporal related issues in the CI pipeline.

Previously the move fixture data was actually being set up for the present day.  Whilst the application technically won't grumble about this it is not how it works in practice.  All move data loaded into the real application is historic with the most recent data being the previous days moves.

Move data is priced on a month by month basis.  Any moves that start on the last day of the month and go into the next month fall within the following months prices.  To avoid this complication and ease integration testing I have changed it so the move fixture data starts on the first day of the previous month and stays within that month.

The integration tests are now steered more towards using the Serco fixture data (though not entirely).  The GEO data is at a fixed point in time which will likely cause issues in the future as new functionality is added (for example preventing historic prices changes when more than two years old).

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ViewAllMoveTypesTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ViewAllMoveTypesTest.kt
@@ -21,20 +21,19 @@ import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMont
 import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMonthMoveData.multiMoveMM1
 import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMonthMoveData.redirectMoveRM1
 import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMonthMoveData.standardMoveSM1
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.previousMonth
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ViewAllMoveTypesTest.kt
```diff
@@ -16,54 +15,60 @@ import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.Pages.Login
...
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMonthMoveData.longHaulMoveLHM1
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMonthMoveData.multiMoveMM1
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMonthMoveData.redirectMoveRM1
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.SercoPreviousMonthMoveData.standardMoveSM1
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.previousMonth
 import java.time.LocalDate
 import java.time.Month
 import java.time.Year
 
-@Disabled
```
matt-crowson-moj, line 29: This was previously disabled due to temporal issue noticed when the month boundary was crossed when the tests were executed on the last day of the month by chance.  This was not an issue with the application, the application works as it should, this was an issue with the fixture data.

Thource (approved):

PR #344, opened 2021-09-16
P4-3006 changes to resolve the highlighting of main menu navigation bar as the user navigates around the application.

matt-crowson-moj:
**Changes:**

The change resolves the highlighting of the main navigation menu bar as the user navigates around the application.  Previously the navigation choice stayed on the same menu item throughout 'Move Summary'.

image

image

image

### DIFF
--- a/src/main/resources/templates/layout/base.html
+++ b/src/main/resources/templates/layout/base.html
@@ -2,134 +2,137 @@
 <html lang="en" class="govuk-template " xmlns:th="http://www.thymeleaf.org"
       xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout">
 
-    <head>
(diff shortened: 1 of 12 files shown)


src/main/resources/templates/layout/base.html
```diff
@@ -2,134 +2,137 @@
...
+
+  appInfoRequest.addEventListener('load', function () {
+    checkAndShowAppInfo(JSON.parse(this.responseText)['message'])
+  })
+
+  appInfoRequest.open('GET', '/app/info');
+  appInfoRequest.send();
+</script>
+</body>
+</html>z
```
thomasleese, line 138:
```suggestion
</html>
```

I think this might be a typo!

thomasleese (dismissed):
I'd suggest we can squash the two commits together, but change looks good!

PR #345, opened 2021-09-16
P4-3127 updates to journeys for review so they are now ordered to make management of them easier for the end user.

matt-crowson-moj:
**Changes:**

On the journeys for review page there was no natural ordering.  This means journeys starting from the same location can be scattered throughout the results on the page.  User feedback has informed us this makes their day to day work harder as they focus on a location at a time.

With these changes results will now be ordered, see attached screenshot as an example:

image

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManageLocationsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManageLocationsTest.kt
@@ -6,6 +6,7 @@ import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestMethodOrder
 import uk.gov.justice.digital.hmpps.pecs.jpc.domain.location.LocationType
 import uk.gov.justice.digital.hmpps.pecs.jpc.domain.price.Supplier
+import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.Journey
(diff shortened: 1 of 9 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManageLocationsTest.kt
```diff
@@ -19,6 +21,7 @@ import uk.gov.justice.digital.hmpps.pecs.jpc.integration.pages.previousMonth
 import java.time.LocalDate
 import java.time.Year
 
+@Disabled
```
matt-crowson-moj, line 24: Last resort really.  I will revisit this but need some time to ponder it.  What I can say is it runs locally just not in the CI pipleline.  Also it all works from a manual point of view.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManageLocationsTest.kt
```diff
@@ -28,7 +31,7 @@ internal class ManageLocationsTest : IntegrationTest() {
 
   @Test
   @Order(1)
-  fun `map missing location name and location type to agency id FROM_AGENCY`() {
+  fun `map missing location name and location type to agency id STOPOVER_AGENCY`() {
```
matt-crowson-moj, line 34: This was an attempt to see if just having one option in the choice of links would fix the CI but alas it did not.

Thource (approved):

PR #349, opened 2021-09-24
P4 3206 price exceptions domain

matt-crowson-moj:
**Changes:**

Introduces the concept of price exceptions at the domain level. This is purely in the backend. As of yet there is no supporting service or front end to add or remove these, this will be added later.

A price exception allows for supplier journey prices to be overridden for a fixed period of a month in a given effective year.  For a given price there can be up to 12 price exceptions in that year (though this is extremely unlikely).

Addition and removal of exceptions is handled by its aggregate root entity 'Price'.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/price/Price.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/price/Price.kt
@@ -19,6 +19,9 @@ import javax.persistence.OneToMany
 import javax.persistence.Table
 import javax.validation.constraints.NotNull
 
+/**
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/price/Price.kt
```diff
@@ -72,6 +83,23 @@ data class Price(
     effectiveYear = effectiveYear,
     addedAt = addedAt
   )
+
+  /**
+   * Duplicate months will be ignored. You must remove first.
```
matt-crowson-moj, line 88: Ignoring duplicates for now.  Not sure but I may change this later e.g. simply replace what is there. Fundamentally though this gives us something to work with for now.

thomasleese (approved):

PR #350, opened 2021-09-24
P4-3214 intiial commit to update repository queries to factor in price exceptions and update code impacted by the change.

matt-crowson-moj:
**Changes:**

The purpose of this change is to use price exceptions in place of prices for a given month if there are any.

It contains updates to the queries that aggregate/pull back prices in the custom SQL (non JPA) repositories and any areas impacted by the change.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/move/JourneyQueryRepository.kt
```diff
@@ -90,19 +90,20 @@ class JourneyQueryRepository(@Autowired val jdbcTemplate: JdbcTemplate) {
                    j.to_nomis_agency_id                                       as journey_to_nomis_agency_id,
                    jtl.site_name                                              as journey_to_site_name,
                    jtl.location_type                                          as journey_to_location_type,
-                   p.price_in_pence                                           as unit_price_in_pence,
+                   COALESCE(pe.price_in_pence, p.price_in_pence)              as unit_price_in_pence,
```
matt-crowson-moj, line 93: Essentially we are saying in these queries a price exception takes precedence if there is one.

thomasleese (approved):

PR #354, opened 2021-09-30
P4-3209 add front end validtion to prevent adding a price exception the same price as the actual price.

matt-crowson-moj:
**Changes:**

Small change to prevent adding price exceptions with the same price as the actual price (as this does not make sense). This is enforced at the Domain level so needs to be enforced here otherwise it will not be handled gracefully.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/MaintainSupplierPricingControllerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/MaintainSupplierPricingControllerTest.kt
@@ -587,6 +587,32 @@ class MaintainSupplierPricingControllerTest(@Autowired private val wac: WebAppli
     verify(service, never()).addPriceException(any(), any(), any(), any(), any(), any())
   }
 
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/controller/MaintainSupplierPricingControllerTest.kt
```diff
@@ -587,6 +587,32 @@ class MaintainSupplierPricingControllerTest(@Autowired private val wac: WebAppli
     verify(service, never()).addPriceException(any(), any(), any(), any(), any(), any())
   }
 
+  @Test
+  internal fun `cannot add an the same price for a  price exception`() {
```
Thource, line 591: `cannot add an exception for the same price?`

matt-crowson-moj: I think a case of more haste less speed, doh! Done.

Thource (approved):

PR #363, opened 2021-10-14
P4 3223 lockdown price changes

matt-crowson-moj:
**Changes**:

Once prices get beyond a certain point they can no longer be changed.  In our case this is when we try and make price changes prior to the previous effective year (so not last year, the year before).  By this time all payments to the suppliers are considered done and dusted and therefore we should not be able to make changes.


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/integration/ManageLocationsAndPricesTest.kt
```diff
@@ -27,85 +39,136 @@ internal class ManageLocationsAndPricesTest : IntegrationTest() {
...
+    isAtPage(JourneyResults)
+      .isAtResultsPageForJourney("PRISON ONE", "POLICE ONE")
+      .isJourneyRowPresent("PRISON ONE", "POLICE ONE", Money.valueOf(50.00))
+      .navigateToUpdatePriceFor("PRISON1", "POLICE1")
+
+    isAtPage(UpdatePrice)
+      .isAtPricePageForJourney("PRISON1", "POLICE1")
+      .assertTextIsNotPresent("Update price")
+      .assertTextIsNotPresent("Price exceptions")
+      .assertTextIsPresent("Price history")
```
matt-crowson-moj, line 149: The key here is we should only see the price history (and not the 'update price' or 'price exceptions') when trying to update an existing price.

PR #364, opened 2021-10-15
P4-3263 updating 3rd party dependencies where available (in preparation for attempt at JDK 17 upgrade).

matt-crowson-moj:


build.gradle.kts
```diff
@@ -69,7 +67,7 @@ dependencies {
 tasks {
   test {
     useJUnitPlatform()
-    exclude("**/integration/*", "**/migrations/*")
```
matt-crowson-moj, line 72: This is actually just removing a reference to a now redundant test package which no longer exists.

PR #376, opened 2021-11-05
P4 3281 single sign on and off

matt-crowson-moj:
**Changes:**

This is the first change of two related to SSO.  This particular change improves the cookie security.

In particular the key changes being setting the http-only attribute and max-age of the cookie.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/interceptors/SessionCookieProperties.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/interceptors/SessionCookieProperties.kt
@@ -1,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.interceptors
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.ConstructorBinding
+import java.time.Duration
+
+@ConstructorBinding
+@ConfigurationProperties(prefix = "server.servlet.session.cookie")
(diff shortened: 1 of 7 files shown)


src/main/resources/application.yml
```diff
@@ -21,6 +21,7 @@ spring:
     store-type: jdbc
     jdbc:
       initialize-schema: never
+    timeout: 20M
```
matt-crowson-moj, line 24: So the default for session timeout is 30 mins.  I have reduced this to 20 mins on the back of the BSI recommendation.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/interceptors/SessionCookieProperties.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.pecs.jpc.interceptors
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.ConstructorBinding
+import java.time.Duration
+
+@ConstructorBinding
+@ConfigurationProperties(prefix = "server.servlet.session.cookie")
+data class SessionCookieProperties(val maxAge: Duration, val httpOnly: Boolean)
```
matt-crowson-moj, line 9: With these annotations (and class) we can easily pull in the matching Spring properties and reference them via an object in the code.

thomasleese (approved):

Thource (approved):

PR #377, opened 2021-11-05
P4-3281 changing how secure attibute is set. Did not work as expexcted when tested over HTTPS connection.

matt-crowson-moj:
**Changes:**

Tweaking how cookie 'secure' attribute is being set.  Did not work as expected even though documentation suggested otherwise.

image

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/interceptors/SessionCookieInterceptor.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/interceptors/SessionCookieInterceptor.kt
@@ -22,7 +22,7 @@ class SessionCookieInterceptor(private val properties: SessionCookieProperties) 
     request.cookies?.firstOrNull { it.name == SPRING_SESSION_COOKIE }?.apply {
       this.maxAge = properties.maxAge.toSeconds().toInt()
       this.path = ANY_PATH_STARTING_AT_ROOT
-      this.secure = request.isSecure
+      this.secure = properties.secure
       this.isHttpOnly = properties.httpOnly
     }?.also { response.addCookie(it) }
   }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/interceptors/SessionCookieInterceptor.kt
```diff
@@ -22,7 +22,7 @@ class SessionCookieInterceptor(private val properties: SessionCookieProperties)
     request.cookies?.firstOrNull { it.name == SPRING_SESSION_COOKIE }?.apply {
       this.maxAge = properties.maxAge.toSeconds().toInt()
       this.path = ANY_PATH_STARTING_AT_ROOT
-      this.secure = request.isSecure
```
matt-crowson-moj, line 25: Documentation says this will be determined by the request type e.g. secure HTTPS.  This did not appear to be the case after looking at the deployment on DEV>

Thource (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/interceptors/SessionCookieInterceptor.kt
thomasleese: Do we think this is perhaps because the load balancer in front of the app server terminates the HTTPS? I wonder if you could configure the app to pick up the value from the (non-standard, but pretty common) [`X-Forwarded-Proto` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)?

matt-crowson-moj: Too be honest (hence the PR) I am not sure.  I didn't investigate that thoroughly.  I am sure the docs are probably right and it is more environmental like you say.

PR #378, opened 2021-11-08
P4 3281 single sign on and off v2

matt-crowson-moj:
**Changes:**

- 'HMPPS' link added which takes user back to main service page.
- The service name is now displayed in the header of the page.
- The 'sign out' now logs the user out of auth as well as the service.

These changes bring it more in line with the BaSM page.

image


matt-crowson-moj:
As discussed I'd question the value of have tests in non UK time zones.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/SecurityConfiguration.kt
```diff
@@ -117,10 +116,12 @@ class SecurityConfiguration<S : Session> : WebSecurityConfigurerAdapter() {
 
       request.session.invalidate()
 
-      response.sendRedirect(authLogoutSuccessUri)
+      response.sendRedirect(ssologoutUri())
     }
   }
 
+  private fun ssologoutUri() = authLogoutSuccessUri.plus("/sign-out")
```
Thource, line 123: <pedantic>I think this should be `ssoLogoutUri`</pedantic>

matt-crowson-moj: Yes it should. Ta.  There is another tweak I would like to make so will sort out then.

PR #381, opened 2021-11-17
P4 3334 price adjustment increase decimal points

matt-crowson-moj:
**Changes:**

- Housekeeping to bump 3rd party dependencies.  This has helped clean up 3rd party CVE transitive dependency overrides.
- Bumping the number of allowed decimals from 4 to 15 for annual price adjustments.  It turns out this level of accuracy is required. 

Whilst there is no financial impact to the change itself, there is a financial impact to the Supplier, as they will not be fairly compensated for the journeys they complete as Indexation would not have been properly applied to the journeys.

image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "3.3.12"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "3.3.13"
   kotlin("plugin.spring") version "1.5.31"
   kotlin("plugin.jpa") version "1.5.31"
   kotlin("plugin.allopen") version "1.5.31"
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/spreadsheet/outbound/PriceSheet.kt
```diff
@@ -1,6 +1,6 @@
 package uk.gov.justice.digital.hmpps.pecs.jpc.service.spreadsheet.outbound
 
-import org.apache.pdfbox.util.Hex
+import org.apache.commons.codec.binary.Hex
```
matt-crowson-moj, line 3: Moving to a later version of Apache POI this was no longer available in this package hence the change.

build.gradle.kts
```diff
@@ -17,54 +17,23 @@ dependencyCheck {
...
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
   implementation("org.springframework.boot:spring-boot-starter-webflux")
-  implementation("org.springframework.session:spring-session-jdbc:2.5.3")
+  implementation("org.springframework.session:spring-session-jdbc:2.6.0")
   implementation("org.springframework.shell:spring-shell-starter:2.0.1.RELEASE")
   implementation("org.thymeleaf.extras:thymeleaf-extras-springsecurity5:3.0.4.RELEASE")
   implementation(kotlin("script-runtime"))
 
-  constraints {
```
matt-crowson-moj, line 53: Hurray, overrides no longer needed. A long awaited update to Apache POI is now available.

thomasleese (approved):

PR #396, opened 2022-01-20
P4-3432 resolves an issue where subsequent events were not being included as part of checking a cancelled move.

matt-crowson-moj:
This is the first part of the fix for this particular bug.  This will ensure going forwards cancelled moves should be handled correctly.

A separate PR (migration??) may be needed to fix historic cancelled moves.  There is an alternative which I am going to weigh up when this is released to preprod, and that is trying the backfilling of reports.


README.md
```diff
@@ -110,12 +110,12 @@ export $(cat .env | xargs) # Only run this if you want to set or update the curr
 ```
 # Import the supplier price spreadsheet that has been uploaded to S3
 
-java -jar app.jar --spring.main.web-application-type=none --price-import --supplier SERCO/GEOAMEY --year 2020
+java -jar app.jar --spring.main.web-application-type=none --price-import --supplier=SERCO/GEOAMEY --year=2020
```
matt-crowson-moj, line 113: Noticed an error in the readme for running the commands.  No related to this PR fix.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/move/MovePersister.kt
```diff
@@ -80,6 +80,8 @@ class MovePersister(
 
           val moveToSave = newMove.copy(
             moveType = moveWithJourneysAndEvents.moveType(),
+            events = moveEvents,
```
matt-crowson-moj, line 83: This was the crux of the problem.  The moveToSave was/is being used to determine the state of a cancelled move which in turn creates a journey so it can be priced and the suppliers paid.

The thing to be aware this does not impact what gets persisted on the actual move as these attributes are transient on the Move entity.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/ImportServiceIntegrationTest.kt
```diff
@@ -48,11 +57,26 @@ internal class ImportServiceIntegrationTest(
...
+  internal fun `given some report files contain errors the import should still complete successfully and not fail`() {
     service.importReportsOn(LocalDate.of(2020, 12, 1))
 
     verify(monitoringService).capture("moves: persisted 1 out of 2 for reporting feed date 2020-12-01.")
     verify(monitoringService).capture("people: persisted 3 out of 4 for reporting feed date 2020-12-01.")
     verifyNoMoreInteractions(monitoringService)
   }
+
+  @Test
+  internal fun `given a booked move which is subsequently cancelled in time, a journey for pricing should be created so the supplier can be paid`() {
```
matt-crowson-moj, line 69: The data used for this was based on the move in question that was raised by the commercial team.  I have removed/changed any identifying data.

matt-crowson-moj, line 69: This is as close as we can get to exercising this for real. The main difference being this is using an in-memory DB.

Thource (approved):

PR #397, opened 2022-01-21
P4-3435 fixing issue parsing historic audit events for price changes.

matt-crowson-moj:
I identified a problem when trying to make a change to a price whereby if there was an historic audit record for the chosen price it can cause a runtime exception.

This is down for a format change.  It was stored as a double but now as a string.

This change adds support for both the old and new format change.


Thource (commented):
LGTM

PR #409, opened 2022-02-02
P4 3467 fix vehicle reg for moves and journeys

matt-crowson-moj:
This change actually removes the registration field from the actual Move entity.  We can derive this from the journeys.

It also addresses an issue in the spreadsheet whereby the vehicle registration for journeys was not being displayed correctly.  You would see the registrations duplicated across all the journeys and as a result would not be able to tell which registration was used for which journey.


src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/move/MoveTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+  @Test
+  fun `given a move with two journeys and different vehicles there should be two vehicle registrations`() {
+    val moveWithTwoJourneys = move().copy(journeys = listOf(journeyOne, journeyTwo))
+
+    assertThat(moveWithTwoJourneys.registration()).isEqualTo("ABC, DEF")
+  }
+
+  @Test
+  fun `given a move with two journeys and the same vehicle there should be one vehicle registration`() {
```
matt-crowson-moj, line 31: I am not 100% sure with this particular change but have decided to go with it for now.

Essentially this means if there was more than one journey with the same registration you would only see one registration for the overall move.  For example 3 journeys with 'ABC' would look like 'ABC' instead of 'ABC, ABC, ABC'.

thomasleese (approved):

PR #412, opened 2022-02-07
P4-3476 split import of people and profiles out from moves to help reduce memory consumption when high volumes of reporting data.

matt-crowson-moj:
The change splits out the import of people and profiles to run separately from that of moves in an effort to reduce memory consumption when there is a large volume of reporting data.

There is also a small amount of refactoring around the existing tasks to tidy them up a bit.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/TaskScheduler.kt
```diff
@@ -6,31 +6,34 @@ import org.springframework.scheduling.annotation.Scheduled
...
+  val previousDaysPeopleAndProfiles: PreviousDaysPeopleAndProfilesTask,
+  val previousDaysLocationMapping: PreviousDaysLocationMappingTask,
+  val backfillReports: BackfillReportsTask
 ) {
 
   @Scheduled(cron = "\${CRON_IMPORT_REPORTS}")
   @SchedulerLock(name = "importReports")
   fun previousDaysReportsTask() {
-    previousDaysReportsTask.execute()
+    previousDaysMoves.execute()
```
matt-crowson-moj, line 24: So splitting out her for now to avoid the need to add another CRON job env var.  This may need to be scheduled to run with a bit of gap if it does not resolve the issue.

Thource (approved):

PR #413, opened 2022-02-08
P4 3476 improve memory consumption usage

matt-crowson-moj:
This changes the importing process for both people and profiles to be sequence (stream) based to help with memory consumption when we have large volumes of data.

This does not get round the fact due to ongoing backfill we have some very large files being processed e.g. 195MB for a profiles file (of which was about 54k individual profiles).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/move/PersonPersister.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/move/PersonPersister.kt
@@ -15,50 +15,48 @@ class PersonPersister(
   /**
    * Returns the total number of successfully persisted people.
    */
-  fun persistPeople(people: List<Person>): Int {
+  fun persistPeople(people: Sequence<Person>): PersistenceResult {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/move/PersonPersister.kt
```diff
@@ -15,50 +15,48 @@ class PersonPersister(
   /**
    * Returns the total number of successfully persisted people.
    */
-  fun persistPeople(people: List<Person>): Int {
+  fun persistPeople(people: Sequence<Person>): PersistenceResult {
```
matt-crowson-moj, line 18: Now that we are using sequences we don't know in advance how many rows there are to be processed hence we need to send back how many were persisted and how many failed to give us a idea of how many rows were actually processed.

thomasleese (approved):
Change looks good! I wonder how much of an impact this will have and whether it'll allow me to increase the size of the batches of journey dates I'm backfilling each day.

PR #422, opened 2022-02-25
P4-3504 changes to prevent potentential redirects going into the standard moves bucket.

matt-crowson-moj:
This is the first attempt (in principle) to resolve an issue whereby we are seeing potential redirections going in to the standard moves category. They are potential redirects as in this situation they are actually redirects missing a journey.  A redirect must have two journeys only. As a result if they are missing journeys they should be showing up in the mutli move category to be resolved with the suppliers and the commercial team.

**_Note: this change will only fix things going forwards.  Will need to run the historic move processing CLI command after this is deployed to fix historic moves._**


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/MoveFilterer.kt
```diff
@@ -45,7 +45,7 @@ object MoveFilterer {
         count { it.stateIsAnyOf(JourneyState.completed) } == 1 &&
           count { isCompleteBillableJourneyAndLocationsMatchMove(it, move) } == 1 &&
           count { it.stateIsAnyOf(JourneyState.cancelled) } == 0
-      }
+      } && move.hasNoneOf(EventType.MOVE_REDIRECT)
```
matt-crowson-moj, line 48: So for now I have kept the change very simple.  We may need to in the future include a time check on the occurred_at time.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/MoveFiltererTest.kt
```diff
@@ -147,6 +147,26 @@ internal class MoveFiltererTest {
...
+        occurredAt = from.atStartOfDay()
+      ),
+      moveEventFactory(
+        type = EventType.MOVE_COMPLETE.value,
+        moveId = "M7",
+        occurredAt = from.atStartOfDay().plusHours(4)
+      )
+    ),
+    journeys = listOf(
+      reportJourneyFactory(journeyId = "J1M6", moveId = "M7", billable = true),
```
matt-crowson-moj, line 166: This is why it goes into multi and not redirect.  It should have two journeys.

Thource (approved):

PR #426, opened 2022-03-01
P4 3504 reverting revert for redirects going into standard moves bucket.

matt-crowson-moj:
This PR was originally reverted as we did not want this to be released before March 9th which marks the end of the supplier backfill.  It is to remain as a draft PR until that date arrives. 

It includes some extra changes on top of the original PR due to gaining a better understanding of the requirements when identifying the specific moves types, which in this case is around standard moves and redirections. The situation being redirects being put into the standard moves box when they had insufficient journeys and should have been going into the multi on the premise the missing journey data will be provided at a later date. Also the redirect event itself should occur after the move has started.

**_Note: when this is merged (and released) the historic moves will need to be corrected.  This can be done via the CLI command 'process-historic-moves'._**


src/main/kotlin/db/migration/data/dev/R__2_5_Integration_test_data.kt
```diff
@@ -327,20 +364,35 @@ class R__2_5_Integration_test_data : BaseJavaMigration() {
...
+              journey(move, fromAgencyId = "FROM_AGENCY", toAgencyId = "TO_AGENCY", state = JourneyState.cancelled),
+              template
+            )
+          )
+        ).failMigrationIfNotMoveType(MoveType.CANCELLED)
       }
     }
   }
 
+  private fun Move.failMigrationIfNotMoveType(moveType: MoveType) {
```
matt-crowson-moj, line 389: This checks a move that has been created is in the correct state.  If not then the migration will fail.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/MoveFilterer.kt
```diff
@@ -36,16 +37,18 @@ object MoveFilterer {
...
   fun isStandardMove(move: Move) =
-    isCompleted(move) &&
+    move.isCompleted() &&
       with(move.journeys) {
         count { it.stateIsAnyOf(JourneyState.completed) } == 1 &&
           count { isCompleteBillableJourneyAndLocationsMatchMove(it, move) } == 1 &&
           count { it.stateIsAnyOf(JourneyState.cancelled) } == 0
-      }
+      } && when (val moveStart = move.mayBeMoveStartDate()) {
+      null -> true
```
matt-crowson-moj, line 49: So I was in two minds here on line 49 whether to return true or false.  Keeping it true means it will work as before.  Returning false may have undesired side effects.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/MoveFilterer.kt
```diff
@@ -36,16 +37,18 @@ object MoveFilterer {
   /**
    * A standard move is a completed move with a single completed journey that is billable, the journey from and to
    * destinations should match that of the move, and no cancelled journeys. To be priced as a standard move, the journey
-   * as well as the move must be completed. There also should be no redirects after the move starts, but shouldn't need
```
matt-crowson-moj, line 39: Turns out this check was actually needed.

Thource (approved):

Thource (approved):

PR #443, opened 2022-03-17
P4-3561 this changes the historic move processing service to include moves that have not been assigned a move type. They were previously ignored.

matt-crowson-moj:
Whilst carrying out move corrections on the preprod and production environments I realised moves without move types were not being included when they should have been.

The historic move processing service has now been changed to now include all moves with and without a move type.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/moves/HistoricMovesProcessingService.kt
```diff
@@ -28,31 +28,29 @@ class HistoricMovesProcessingService(
 ) {
 
   /**
-   * This will go through and only process historic moves that have been assigned a move type.
```
matt-crowson-moj, line 31: This was a mistake on my part.

thomasleese (approved):

PR #469, opened 2022-05-16
P4 3664 reconciliation moves feature

matt-crowson-moj:
This change introduces a new tab 'Reconciliation Moves' on the generated pricing spreadsheet for helping identify candidate completed moves which cannot be priced due to missing information e.g. move type or the date and time the move completed. The intent is this can be used in conjunction with the invoices the suppliers provide us.

As a first cut/run it is feature toggled and will only run as high as preprod.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/domain/move/MoveRepository.kt
```diff
@@ -1,11 +1,31 @@
...
+
+  @Query(
+    """
+    select m
+      from Move m
+     where m.supplier = :supplier
+       and m.moveYear = :year
+       and m.moveMonth = :month
+       and m.status= 'completed'
+       and (m.moveType is null or m.dropOffOrCancelledDateTime is null)
```
matt-crowson-moj, line 22: if a move has been classified as completed but either the move type cannot be established or the drop of date is null they are not included in the overall pricing therefore not visible to the end user.

Thource (approved):

PR #472, opened 2022-05-27
P4-3700 resolving deprecation that arose as a result of 3rd party dependency updates.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/security/SecurityConfiguration.kt
```diff
@@ -37,7 +37,7 @@ private val logger = loggerFor<SecurityConfiguration<*>>()
 @EnableWebSecurity
 @ConditionalOnWebApplication
 @EnableGlobalMethodSecurity(prePostEnabled = true)
-class SecurityConfiguration<S : Session> : WebSecurityConfigurerAdapter() {
```
matt-crowson-moj, line 40: `WebSecurityConfigurerAdapter` is deprecated after moving to a later version of Spring.

Thource (approved):

PR #476, opened 2022-06-16
P4 3722 auto repair report ingress feature

matt-crowson-moj:
This feature change adds support for the service to attempt to auto repair the reports ingress should it fail due to missing reports files

In short it will now go back to the day after the last successful import and use that as a starting point and continue to the day before date of execution.

It has a env var toggle so we can switch it off should we need e.g. genuine reason for it to missing data on a give date.


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/tasks/ImportReportsDataTask.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
+import uk.gov.justice.digital.hmpps.pecs.jpc.service.MonitoringService
+import uk.gov.justice.digital.hmpps.pecs.jpc.service.reports.ImportReportsService
+import java.time.LocalDate
+
+@Component
+class ImportReportsDataTask(
+  @Value("\${IMPORT_REPORTS_BACKDATE_ENABLED}") private val backDateEnabled: Boolean = false,
```
matt-crowson-moj, line 12: This task replaces the two separate tasks for previous days moves, events, journeys task and previous days profiles and peoples task.

helm_deploy/calculate-journey-variable-payments/templates/_envs.tpl
```diff
@@ -145,4 +145,7 @@ env:
   - name: FEATURE_FLAG_INCLUDE_RECONCILIATION_MOVES
     value: "{{ .Values.env.FEATURE_FLAG_INCLUDE_RECONCILIATION_MOVES }}"
 
+  - name: IMPORT_REPORTS_BACKDATE_ENABLED
```
matt-crowson-moj, line 148: So it occurred to me we need to switch this off as there may well situations where we miss days e.g. a major service outage beyond a day whereby we resort back to a paper based system ergo potentially no electronic move data.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/config/tasks/TasksSchedulingConfiguration.kt
```diff
@@ -26,8 +24,7 @@ class TasksSchedulingConfiguration(
   @Scheduled(cron = "\${CRON_IMPORT_REPORTS}")
   @SchedulerLock(name = "importReports")
   fun previousDaysReportsTask() {
-    previousDaysMoves.execute()
-    previousDaysPeopleAndProfiles.execute()
+    importReportsDataTask.execute()
```
matt-crowson-moj, line 27: We now only have one task responsible for importing all data.  This is to make sure all files are present prior to importing.  When split into two the two tasks do not know about each other.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/tasks/ImportReportsDataTask.kt
```diff
@@ -0,0 +1,37 @@
...
+          return
+        }
+      }
+
+      service.importAllReportsOn(this)
+    }
+  }
+
+  private fun backdateReportsFrom(from: LocalDate) {
+    from.datesUntil(timeSource.date()).forEach { service.importAllReportsOn(it) }
```
matt-crowson-moj, line 35: Basically repeats until the present day (the present day is exclusive here, it does not run for this date).

Thource (approved):

PR #479, opened 2022-06-21
P4-3722 resolves an issue where by was not starting at the correct date when resuming from a failed report ingress.

matt-crowson-moj:
Testing in development highlighted a gap in the auto recovery from a failed ingress. Missing import on date 2022-06-18.

image

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/tasks/ImportReportsDataTask.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/tasks/ImportReportsDataTask.kt
@@ -18,10 +18,10 @@ class ImportReportsDataTask(
   override fun performTask() {
     timeSource.yesterday().run {
       if (backDateEnabled) {
-        val importDate = service.dateOfLastImport()?.plusDays(1) ?: this
+        val startDate = service.dateOfLastImport() ?: this
 
-        if (importDate.isBefore(this)) {
-          backdateReportsFrom(importDate)
+        if (startDate.isBefore(this)) {
+          backdateReportsFrom(startDate)
 
           return
         }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/tasks/ImportReportsDataTask.kt
```diff
@@ -18,10 +18,10 @@ class ImportReportsDataTask(
   override fun performTask() {
     timeSource.yesterday().run {
       if (backDateEnabled) {
-        val importDate = service.dateOfLastImport()?.plusDays(1) ?: this
```
matt-crowson-moj, line 21: This is the culprit.  Should not be adding anything here due to the actual reports being offset by a day anyway.  What this means is if the last import was on the 17th then the actual reports imported would be for the 16th.  So when we recover we would start on the 17th, the last date of a successful import.

src/test/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/tasks/ImportReportsDataTaskTest.kt
```diff
@@ -28,8 +29,10 @@ class ImportReportsDataTaskTest {
     ImportReportsDataTask(true, importReportsService, timeSource, monitoringService).execute()
 
     verify(importReportsService).dateOfLastImport()
+    verify(importReportsService).importAllReportsOn(LocalDate.of(2022, 6, 12))
     verify(importReportsService).importAllReportsOn(LocalDate.of(2022, 6, 13))
     verify(importReportsService).importAllReportsOn(LocalDate.of(2022, 6, 14))
+    verifyNoMoreInteractions(importReportsService)
```
matt-crowson-moj, line 35: This ensures all interactions are covered i.e. not missing an assertion.

Thource (approved):

PR #485, opened 2022-07-21
P4-3764 initial commit to add the daily imports runner and associated profile which will be used by the k8 job.

matt-crowson-moj:
Prep work to for the upcoming k8 job so can run the daily reports import upon starting the container based application which will get pulled down as part of the job.

Essentially what we are going to end up with is this in the job:

`java -jar app.jar --spring.profiles.active=daily-reports`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/cli/CommandRunner.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/cli/CommandRunner.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.pecs.jpc.cli
 import org.springframework.boot.ApplicationArguments
 import org.springframework.boot.ApplicationRunner
 import org.springframework.boot.autoconfigure.condition.ConditionalOnNotWebApplication
+import org.springframework.context.annotation.Profile
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.pecs.jpc.domain.price.Supplier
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/cli/DailyReportsImporterRunner.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import org.springframework.boot.ApplicationArguments
+import org.springframework.boot.ApplicationRunner
+import org.springframework.boot.autoconfigure.condition.ConditionalOnNotWebApplication
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.pecs.jpc.config.TimeSource
+import uk.gov.justice.digital.hmpps.pecs.jpc.service.reports.ImportReportsService
+
+@ConditionalOnNotWebApplication
```
matt-crowson-moj, line 11: Essentially this will run on startup of the application when the given profile is used.

src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/cli/CommandRunner.kt
```diff
@@ -15,6 +16,7 @@ import java.time.LocalDate
 private val logger = loggerFor<CommandRunner>()
 
 @ConditionalOnNotWebApplication
+@Profile("!daily-reports")
```
matt-crowson-moj, line 19: This is saying don't run this on application startup when this profile is used.

Thource (approved):

PR #492, opened 2022-08-02
P4-3764 only audit reports import if move, people and profiles imported without error. This indicates a successful import.

matt-crowson-moj:
This should help resolve a self repair issue whereby an import didn't work fully e.g. it only imported moves and people and not profiles.  Previously the service regarded this as successful due to audit records existing for the moves and people imported.  Now the audit events for an import will only be created if all 3 are successfully imported so any attempts for the service to self repair will work from the date of the last successful audit entries.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/ImportReportsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/ImportReportsService.kt
@@ -44,9 +44,17 @@ class ImportReportsService(
     failIfDateOfImportNotInPast(date)
     failIfAnyReportFilesAreMissingOn(date)
 
-    importMovesJourneysEventsOn(date)
-    importPeopleOn(date)
-    importProfilesOn(date)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/ImportReportsService.kt
```diff
@@ -44,9 +44,17 @@ class ImportReportsService(
...
     failIfAnyReportFilesAreMissingOn(date)
 
-    importMovesJourneysEventsOn(date)
-    importPeopleOn(date)
-    importProfilesOn(date)
+    val maybeMovesAuditableEvent = importMovesJourneysEventsOn(date)
+    val maybePeopleAuditableEvent = importPeopleOn(date)
+    val maybeProfilesAuditableEvent = importProfilesOn(date)
+
+    if (maybeMovesAuditableEvent != null && maybePeopleAuditableEvent != null && maybeProfilesAuditableEvent != null) {
```
matt-crowson-moj, line 51: An option here would be to put this into one event instead of 3 which may be cleaner.

Thource (approved):

PR #493, opened 2022-08-02
P4-3764 only audit reports import if move, people and profiles imported without error. This indicates a successful import.

matt-crowson-moj:
This should help resolve a self repair issue whereby an import didn't work fully e.g. it only imported moves and people and not profiles. Previously the service regarded this as successful due to audit records existing for the moves and people imported. Now only one audit event for an import will only be created if all 3 are successfully imported so any attempts for the service to self repair will work from the date of the last successful audit entry.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/ImportReportsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/ImportReportsService.kt
@@ -44,9 +44,23 @@ class ImportReportsService(
     failIfDateOfImportNotInPast(date)
     failIfAnyReportFilesAreMissingOn(date)
 
-    importMovesJourneysEventsOn(date)
-    importPeopleOn(date)
-    importProfilesOn(date)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/pecs/jpc/service/reports/ImportReportsService.kt
```diff
@@ -44,9 +44,23 @@ class ImportReportsService(
...
 
-    importMovesJourneysEventsOn(date)
-    importPeopleOn(date)
-    importProfilesOn(date)
+    val maybeMoves = importMovesJourneysEventsOn(date)
+    val maybePeople = importPeopleOn(date)
+    val maybeProfiles = importProfilesOn(date)
+
+    if (maybeMoves != null && maybePeople != null && maybeProfiles != null) {
+      auditService.create(
```
matt-crowson-moj, line 52: Decided to squash this into 1 audit event as this feels cleaner in the long run.

Thource (approved):

georgeschena (approved):

PR #517, opened 2022-11-14
P4 3921/bulk upload overwrites

danbenton-mojdt:


scripts/perform_annual_adjustment/perform_annual_adjustment.sh
```diff
@@ -0,0 +1,99 @@
...
+ENV=$1
+SUPPLIER=$2
+YEAR=$3
+INFLATIONARY=$4
+VOLUMETRIC=$5
+
+
+if [ -z "$ENV" ]
+then
+  echo "No environment specified, please supply the environment in the first argument, dev, preprod or prod."
```
Thource, line 17: Would be nice to check that $ENV is `dev`, `preprod` or `prod` I think?

danbenton-mojdt: Could do, but if something other than those are provided it can't find any pods to run on so the user can't progress anyway. Would be nice to quit quite verbosely though admittedly. Will add this on another pr to add it to all of the similar scripts

PR #544, opened 2024-06-18
MAP-1280 Install postgres cert

danbenton-mojdt:


Mjwillis (approved):
Does this talk to a db?

danbenton-mojdt:
> Does this talk to a db?

It does, it has its own postgres db in which it stores (transformed) basm data along with journey prices

PR #561, opened 2024-12-24
MAP-1998 fixing application-insights configuration

michalnawrockidj:


helm_deploy/calculate-journey-variable-payments/templates/_envs.tpl
```diff
@@ -94,12 +94,6 @@ env:
         name: {{ template "app.name" . }}
         key: HMPPS_AUTH_CLIENT_SECRET
 
-  - name: APPINSIGHTS_INSTRUMENTATIONKEY
```
marcusaleman, line 97: so if you remove that from the _envs.tpl, how you get the value injected?

marcusaleman: I see in the previous PR set up in APPLICATIONINSIGHTS_CONNECTION_STRING . so I do LGTM

marcusaleman (approved):

PR #564, opened 2025-01-03
MAP-1998 fixing APPINSIGHTS

michalnawrockidj:


marcusaleman (approved):
LGTM as I am following Slack discussion, it looks as discussed temporary sol.

PR #608, opened 2025-06-26
MAP-2535 GitHub deployment actions

michalnawrockidj:


.nvmrc
dafhoustonmoj, line 1: should this be here?

michalnawrockidj: Yes It is needed

dafhoustonmoj (approved):

PR #668, opened 2026-05-19
MAPB-404: replace snyk with trivy and docker compose fix

dafhoustonmoj:


Thource (approved):
Title is the wrong way around :D

PR #683, opened 2026-06-18
MAPB-450: fix oauth2 login configuration

dafhoustonmoj:
**Summary**
The Spring Boot upgrade (HMPPS Gradle plugin 8.x → 10.x) brought in Spring Boot 4.0 / Spring Security 7.0, which introduced two breaking changes to the OAuth2 login flow:
 
**Breaking change 1:** 
PKCE now always enabled (root cause of sign-out)
Spring Security 7.0 unconditionally adds PKCE (code_challenge / code_verifier) to all authorization_code flows. HMPPS Auth rejects confidential clients using PKCE with invalid_client, causing the login to fail and redirect to sign-out.
**Fix:** 
Custom DefaultOAuth2AuthorizationRequestResolver in SecurityConfiguration.kt that strips the PKCE parameters from the authorization request after SS7 adds them.
 
**Breaking change 2:** 
autoconfigure: exclude (red herring, reverted)
An earlier attempt incorrectly excluded HmppsResourceServerConfiguration. This wasn't the root cause and has been removed.


GurnankCheema (approved):
LGTM

